PR #3, opened 2020-04-20
Radio buttons

illright:


src/radio-button/radio-group.svelte
```diff
@@ -0,0 +1,53 @@
...
+  if (!items || items.length === 0) {
+    console.error('Must have at least one item in the radio group');
+  }
+
+  if (color && labelClass != null) {
+    console.warn('labelClass has no effect: labels are not rendered for color radio groups');
+  }
+</script>
+
+{#if items != null && items.length !== 0}
```
aabounegm, line 27: perhaps the div should be rendered anyways but with empty content? Also, I find this check useless because `items` is required, so a wrong value **should** throw an error, and `{#each }` doesn't mind an empty array, so the second part of the check is also meaningless

illright: Don't think it makes much of a difference as to what happens when an erroneous value was given. As for the two checks, it's preventing the each from trying to iterate over null, which produces another error in addition to the clean one produced by our code. Just makes it harder to concentrate on the problem.

PR #4, opened 2020-04-21
Add Card component

aabounegm:
Image has been simplified as just a prop (instead of having to manually apply the `.image` class on it in the default slot). Title and subtitle have also been added as props

### DIFF
--- a/src/card/card.scss
+++ b/src/card/card.scss
@@ -11,26 +11,4 @@
   &.no-padding {
     padding: 0;
   }
-
-  &.with-image {
-    padding: 0;
-
-    & > .image {
-      width: 100%;
-    }
-
-    & > .content {
-      padding: .8em 1.3em;
-    }
-
-  }
-  .title {
-    font-size: 1.15rem;
-    font-weight: 400;
-  }
-
-  .subtitle {
-    font-size: .95rem;
-    font-weight: 300;
-  }
 }
--- a/src/card/card.scss
+++ b/src/card/card.scss
@@ -4,7 +4,7 @@
 .card {
   box-shadow: $shadow-0;
   background-color: #fff;
-  border-radius: 1.25rem;
+  border-radius: $card-radius;
   padding: 1.5em;
   overflow: hidden;
 
--- a/src/card/card.scss
+++ b/src/card/card.scss
@@ -8,6 +8,10 @@
   padding: 1.5em;
   overflow: hidden;
 
+  &.no-padding {
+    padding: 0;
+  }
+
   &.with-image {
     padding: 0;
 
--- a/src/card/card.scss
+++ b/src/card/card.scss
@@ -1,0 +1,32 @@
+@import '../var-defaults.scss';
+@import '_attractions-theme.scss';
+
+.card {
+  box-shadow: $shadow-0;
+  background-color: #fff;
+  border-radius: 1.25rem;
+  padding: 1.5em;
+  overflow: hidden;
+
+  &.with-image {
+    padding: 0;
+
+    & > .image {
+      width: 100%;
+    }
+
+    & > .content {
+      padding: .8em 1.3em;
+    }
+
+  }
+  .title {
+    font-size: 1.15rem;
+    font-weight: 400;
+  }
+
+  .subtitle {
+    font-size: .95rem;
+    font-weight: 300;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/card/card.scss
```diff
@@ -0,0 +1,32 @@
...
+@import '_attractions-theme.scss';
+
+.card {
+  box-shadow: $shadow-0;
+  background-color: #fff;
+  border-radius: 1.25rem;
+  padding: 1.5em;
+  overflow: hidden;
+
+  &.with-image {
```
illright, line 11: And the simplicity of the component also means getting rid of those styles. The only useful prop (and class, for that matter) to my mind is the one that would disable the internal padding.

src/card/card.scss
```diff
@@ -0,0 +1,32 @@
+@import '../var-defaults.scss';
+@import '_attractions-theme.scss';
+
+.card {
+  box-shadow: $shadow-0;
+  background-color: #fff;
+  border-radius: 1.25rem;
```
illright, line 7: Maybe have a variable, `$card-radius`? Seems like squarish cards are quite popular too

src/card/card.scss
```diff
@@ -0,0 +1,32 @@
+@import '../var-defaults.scss';
+@import '_attractions-theme.scss';
+
+.card {
+  box-shadow: $shadow-0;
+  background-color: #fff;
```
illright, line 6: And this, this also asks for a variable. Let it be `$card-bg`

illright: For a second thought, no, let's keep it white. I see no use in changing the card color for all cards, and individual cards can be styled on demand.

src/card/card.svelte
```diff
@@ -0,0 +1,28 @@
+<script>
+  import classes from '../_utils/classes.js';
+
+  let _class = null;
+  export { _class as class };
+  export let id = '';
+  export let imgSrc = '';
```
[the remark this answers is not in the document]
aabounegm: The only point in taking images into consideration is that everything inside a card is padded, but images wouldn't look good. The design we are using for the product card in Innopoints wouldn't be possible otherwise

src/card/card.scss
aabounegm: I don't think people would commonly change the background color of individual cards, but would rather want to change all cards to conform to a certain design (just like with the radius, for example).

src/card/card.svelte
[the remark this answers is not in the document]
aabounegm: I also think it is not uncommon that people would add at least a title to the card. It does no harm, just one line for convenience

src/card/card.scss
aabounegm: In my point of view, libraries should do as much heavy lifting and provide as much functionality as possible out-of-the-box, while not disallowing any customization. Excessive simplicity reduces the added value and increases the amount of work the user should do. However, that is just my personal view; or are we going with a very simplistic and minimalist design for the library?

illright: The truth is, the reason why I don't want those things in the code is because they are heavily tailored to innopoints. There's actually a lot of conventions regarding cards, but this card barely satisfies the image one. So it's either we remove all those styles or rewrite the card properly. And I'm not up for the latter

src/card/card.svelte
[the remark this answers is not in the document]
illright: A title to the card? Never seen anything like it. It does bring clutter and possibly harm Svelte optimizations. I'd say we remove it, for I really see no purpose for it.

src/card/card.scss
illright: So since you added that `tight` prop, the rest of the contents of the Card are no longer necessary in my opinion

aabounegm: I'm not sure what "properly" exactly means in this context, or what the conventions are, so I'll just go with the first option then.

illright:
What about `restProps`?

aabounegm:
> What about `restProps`?

It is there
https://github.com/illright/attractions/pull/4/files#diff-9a80d42b524018c9444bcb672e94223aR9

illright:
Yeah, I'm saying we should go without it. Check the unresolved discussions

aabounegm:
I took the "I'd say we could do a restProps" part of the sentence. And how about assigning id or any of the ARIA attributes then?
In Vue, attributes not defined in props automatically fall through to the root components, and this seemed to be what you did with the other components

PR #5, opened 2020-04-22
Feature/demo page

aabounegm:
Add a page for showcasing the components. Didn't design anything properly, so it is just a proof-of-concept. Wrote a GitHub Action to deploy that demo page to GitHub Pages, but cannot test it without first making a release.


demo/App.svelte
```diff
@@ -0,0 +1,35 @@
+<script>
+	import {
```
illright, line 2: Tabs

jsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+	"compilerOptions": {
+		"baseUrl": "./",
+		"paths": {
+			"@/*": ["src/*"],
```
illright, line 5: You sure this is necessary? It was introduced to Innopoints by Max and I never really saw the need in it, it's confusing

package.json
```diff
@@ -8,25 +8,33 @@
...
   "keywords": [
     "svelte"
   ],
   "files": [
     "src",
     "dist"
-  ]
+  ],
+  "dependencies": {
+    "sirv-cli": "^0.4.5"
```
illright, line 38: That definitely should not be a dependency

jsconfig.json
aabounegm: It helps avoid writing a lot of '../../' imports. Also, this particular file does nothing but tell VS Code about that alias. The actual alias is defined in the rollup config

illright: But in this case it only substitutes `@/` for `src/`, why couldn't we just write `src/` directly?

PR #12, opened 2020-05-05
Checkboxes

illright:


src/checkbox/checkbox-group.svelte
```diff
@@ -0,0 +1,64 @@
...
+
+  export let color = false;
+
+  export let items;
+  export let name;
+  export let labelsLeft = false;
+  export let max = null;
+
+  $: currentChecked = items.reduce((acc, elt) => acc + elt.checked, 0);
+  const maxReachedTooltip = { title: `Can only select ${max} value${s(max)}.` };
```
aabounegm, line 20: Perhaps this should be a prop to allow for better customization and i18n.

illright: Will add as a future improvement

PR #14, opened 2020-05-08
Chips

illright:


src/chip/checkbox-chip-group.svelte
```diff
@@ -0,0 +1,51 @@
...
+        class={checkboxClass}
+        on:change
+        {...$$restProps}
+        {...(
+          !item.disabled
+          && !item.checked
+          && max != null
+          && currentChecked >= max
+          ? maxReachedTooltip : {}
+        )}
```
aabounegm, line 39: How about moving this whole condition into an appropriately-named reactive variable and using the `title` prop directly instead of the weird spread?
Also, doesn't having it after the `$$restProps` spread make it override it? In that case, it would be better to keep it before it so the user can override the tooltip manually (until better i18n support and so on)

src/chip/radio-chip.svelte
```diff
@@ -0,0 +1,45 @@
...
+  <div
+    class:small
+    class:outline
+    class:no-padding={noPadding}
+    {title}
+    use:ripple={{ disabled }}
+    class={classes('chip', chipClass)}
+  >
+    <slot>{value}</slot>
+  </div>
```
aabounegm, line 42: Why doesn't this use the Chip component? It seems almost the same

src/chip/chip.svelte
```diff
@@ -0,0 +1,15 @@
+<script>
+  import classes from '../_utils/classes.js';
+
+  let _class = null;
+  export { _class as class };
+  export let small = false;
+  export let outline = false;
+  export let noPadding = false;
```
aabounegm, line 8: I feel like allowing to specify a custom padding value would be more flexible
```suggestion
  export let padding = 0;
```
and then use in `style="padding: {padding}px;"` or something similar

src/chip/checkbox-chip.svelte
```diff
@@ -0,0 +1,43 @@
...
+</script>
+
+<label class={classes('chip-wrapper', _class)}>
+  <input
+    bind:checked
+    {value}
+    type="checkbox"
+    class={inputClass}
+    {disabled}
+    on:change={(e) => dispatch('change', { value: e.target.value, checked: e.target.checked })}
```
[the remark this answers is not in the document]
illright: Then I suppose we had a misunderstanding. All events emitted from components are actually custom, with only relevant properties at quick access.

src/chip/chip.svelte
illright: I'm psychologically against inline styles. Ew-ew

src/chip/radio-chip.svelte
illright: Ripples

aabounegm: What if you use `...$$restProps` in the chip component?

src/chip/chip.svelte
aabounegm: Well, since Svelte doesn't support using JS variables in CSS, I see no other way.

src/chip/radio-chip.svelte
illright: Actions are not props, that wouldn't work

src/chip/chip.svelte
illright: Except for the way this is built now) one can still style the cards with CSS

src/chip/checkbox-chip.svelte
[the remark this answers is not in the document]
aabounegm: ooh, that was a lengthy discussion). I don't see on what basis did you decide that these are the only "relevant" properties, while in the same time wanting to use an array of event listeners to give full power and flexibility to the user and not decide by yourself the relevant events.
I believe that giving the full event object will be most flexible as there are often other properties needed for better control (for example, preventing default). It would be better if the native event is augmented with the relevant properties for quick access
```suggestion
    on:change={(e) => dispatch('change', Object.assign(e, { value: e.target.value, checked: e.target.checked }))}
```
or just sent along with them
```suggestion
    on:change={(e) => dispatch('change', { event: e, value: e.target.value, checked: e.target.checked })}
```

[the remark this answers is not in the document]
illright: This is reasonable. I like the second option more. But the same issue then needs to be fixed in all components, so I'll create an issue for this and make it a separate PR.

PR #15, opened 2020-05-12
Switches

illright:


src/switch/switch.svelte
```diff
@@ -0,0 +1,38 @@
...
+
+  let _class = null;
+  export { _class as class };
+  export let inputClass = null;
+  export let trackClass = null;
+  export let thumbClass = null;
+
+  export let value = false;
+  export let disabled = false;
+  export let slotLeft = false;
```
aabounegm, line 13: I feel like the slot will almost always be used just for a label, so I feel like the name `labelLeft` would be more appropriate.

src/switch/switch.svelte
```diff
@@ -0,0 +1,38 @@
...
+<label class={classes('switch', _class)}>
+  {#if slotLeft}
+    <slot />
+  {/if}
+  <input
+    bind:checked={value}
+    type="checkbox"
+    class={inputClass}
+    {disabled}
+    on:change={(e) => dispatch('change', { value: e.target.checked })}
```
aabounegm, line 27: Same comment as in #14 about dispatched events
```suggestion
    on:change
```

src/switch/switch.svelte
illright: This is kind of for consistency with Checkbox and RadioButton, which also have a `slotLeft` prop, while CheckboxGroup has `labelsLeft`. In that way, the Checkbox and RadioButton components are meant to be used sparingly, where extra customization is needed, and in that case the `slotLeft` name wouldn't scare people away. As for the switch, again, consistency.

src/switch/switch.svelte
illright: Same thing, to be fixed later

PR #17, opened 2020-05-16
Tabs

illright:


src/tab/tab.svelte
```diff
@@ -0,0 +1,34 @@
...
+  export let group = null;
+  export let disabled = false;
+  export let ripple = true;
+
+  const dispatch = createEventDispatcher();
+</script>
+
+<label class={classes('tab', _class)}>
+  <input
+    bind:group
```
aabounegm, line 21: I cannot find the attribute named `group` anywhere in the [MDN docs](https://www.youtube.com/watch?v=rNr_0TUXmAA). Did you mean `name`?

src/tab/tab.svelte
```diff
@@ -0,0 +1,34 @@
...
+    {value}
+    type="radio"
+    class={inputClass}
+    {disabled}
```
aabounegm, line 25: Isn't it redundant to include it as a prop when it can just be passed in `$$restProps` since it is not used anywhere else? (unlike `value`, which is used elsewhere.)

src/tab/tab.svelte
illright: No, this is Svelte-specific. This would be the value of the currently selected radiobutton

illright: Nice link, btw :p

src/tab/tab.svelte
illright: I suppose I just wanted to document this prop to let people know that this is the way to disable tabs

src/tab/tab.svelte
aabounegm: Oops, Windows clipboard did it again 😅. Fixed link.
Not sure why Svelte did this when we have the `name` attribute, but ok

src/tab/tab.svelte
aabounegm: Documenting can be done in the documentation. I personally don't check out the source code unless I meet some weird behavior I need to explain

illright: Yeah, but it doesn't feel right to document a prop that isn't actually there.

src/tab/tab.svelte
illright: The `name` attribute serves a different purpose. `name` defines a group of radiobuttons, while the `group` attribute contains a value of the currently selected radiobutton, the same for all radiobuttons.

PR #18, opened 2020-05-16
Pass the native event object along where applicable

### LINKED ISSUE #16: Pass the native event object in the detail of some custom events
For instance, a selection of a chip should have the following detail object: `{ value, checked, nativeEvent }`.

The rename of `event` to `nativeEvent` is made to make the `event.detail.event` access chain less awkward.

illright:
Closes #16

### DIFF
--- a/src/chip/radio-chip.svelte
+++ b/src/chip/radio-chip.svelte
@@ -27,7 +27,7 @@
     type="radio"
     class={inputClass}
     {disabled}
-    on:change={() => dispatch('change', value)}
+    on:change={() => dispatch('change', { value, nativeEvent: e })}
     {...$$restProps}
   />
   <div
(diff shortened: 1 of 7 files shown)


src/chip/radio-chip.svelte
```diff
@@ -27,7 +27,7 @@
     type="radio"
     class={inputClass}
     {disabled}
-    on:change={() => dispatch('change', value)}
+    on:change={() => dispatch('change', { value, nativeEvent: e })}
```
aabounegm, line 30: Where is `e` coming from? You forgot to capture it in the argument list

src/radio-button/radio-button.svelte
```diff
@@ -26,7 +26,7 @@
     type="radio"
     class={inputClass}
     {disabled}
-    on:change={() => dispatch('change', value)}
+    on:change={() => dispatch('change', { value, nativeEvent: e })}
```
aabounegm, line 29: Again, `e` not defined

PR #22, opened 2020-05-18
Dots

illright:


src/dot/dot.svelte
```diff
@@ -0,0 +1,28 @@
...
+  export { _class as class };
+
+  export let info = false;
+  export let attention = false;
+  export let danger = false;
+  export let success = false;
+
+  export let small = false;
+
+  if (info + attention + success > 1) {
```
aabounegm, line 14: Why not checking for `danger` as well?

illright: Because I forgot)

illright:
Yeah, that'd require `style`.

PR #24, opened 2020-05-18
Typography

illright:


demo/App.svelte
```diff
@@ -31,7 +33,10 @@
   let colorItem = colorItems[1].value;
 </script>
 
+<a href="https://googles.com">say my name</a>
```
aabounegm, line 36: Not sure if you meant Google, but this website seems cool as well 😆

illright: Nope, I deliberately wanted something that my browser didn't visit before)

illright:
> I would have preferred if they were just global classes people can apply to anything and not actual useless components, although I like the idea of enforcing semantics.

The idea of global classes doesn't go well with the approach of scoped component CSS philosophy. That might have been more flexible, but let's be consistent. And semantics, yeah.

> Also, why support only h1 through h3 while HTML has up to h6?

Technically, we only have styles for up to 3 headings, the rest are not frequent at all, and can be styled on demand by the library users themselves. Especially to discourage those (like the past me) who use only `<h5>` and `<h6>` just because they are __small__

PR #25, opened 2020-05-18
Accordions

illright:


src/accordion/accordion-section.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import classes from '../_utils/classes.js';
+
+  let _class = null;
+  export { _class as class };
+
+  export let label = null;
+  export let open = false;
+
+  const selfControl = {
+    close() {
```
aabounegm, line 13: So there is `close` and `toggle` but not `open`?

src/accordion/accordion-section.svelte
```diff
@@ -0,0 +1,37 @@
...
+      open = false;
+    },
+    toggle() {
+      if (!open) {
+        dispatch('panel-open', selfControl);
```
aabounegm, line 18: Why not dispatching when it is closed as well? Also, where is the native event?

src/accordion/accordion-section.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import { createEventDispatcher } from 'svelte';
+  import classes from '../_utils/classes.js';
+
+  let _class = null;
+  export { _class as class };
+
+  export let label = null;
+  export let open = false;
+
+  const selfControl = {
```
aabounegm, line 12: I don't really see the need for an object here. It's not like we have a lot of them and need to organize internal states separately

src/accordion/accordion-section.svelte
illright: This `selfControl` is a piece of internal API anyway, and only `toggle` is exposed as a slot prop, so no problem here. `open` is not used internally.

src/accordion/accordion-section.svelte
illright: The need for an object is simply for guaranteed identification of each panel. Objects have referential equality, therefore, each panel is uniquely identified by its `selfControl` object, which is the behaviour that `<Accordion>` relies on

src/accordion/accordion-section.svelte
illright: Technically, this is an internal event that wasn't really meant to be listened on, but now that I think about it, why not. Where's the native event, you ask? There might not be. We don't know what has triggered the toggle of the panel, it's just a callback that a user calls at their liking

demo/App.svelte
```diff
@@ -36,9 +38,9 @@
   <Button filled>
     Button!!
   </Button>
-  <RadioGroup items={colorItems} color bind:value={colorItem} name="colors" />
+  <!-- <RadioGroup items={colorItems} color bind:value={colorItem} name="colors" />
   <RadioGroup items={items} bind:value={item} name="numbers" />
-  <CheckboxGroup items={items} max={1} name="numbers-check" />
+  <CheckboxGroup items={items} max={1} name="numbers-check" /> -->
```
aabounegm, line 43: Why is this commented out?

illright: It was too long on the page and making me sick. This isn't even a proper demo page anyway, I just needed some space to showcase the components from the latest PRs. It's okay to leave this commented.

PR #26, opened 2020-05-20
Snackbars

illright:
Svelte is such a superpower

### DIFF
--- a/src/snackbar/snackbar-container.svelte
+++ b/src/snackbar/snackbar-container.svelte
@@ -15,7 +15,7 @@
     const {
       component = Snackbar,
       props = {},
-      duration = 3000,
+      duration = 4000,
     } = options;
 
     const key = { component, props };
--- a/src/snackbar/snackbar-container.svelte
+++ b/src/snackbar/snackbar-container.svelte
@@ -25,8 +25,7 @@
 
     return function close() {
       clearTimeout(key.timeoutID);
-      registeredSnackbars.delete(key);
-      registeredSnackbars = registeredSnackbars;
+      removeSnackbar(key);
     };
   }
 
--- a/src/snackbar/snackbar-container.svelte
+++ b/src/snackbar/snackbar-container.svelte
@@ -1,0 +1,46 @@
+<script>
+  import { setContext } from 'svelte';
+  import Snackbar from './snackbar.svelte';
+  import snackbarContextKey from './snackbar-context-key.js';
+
+  export let position;
+  let registeredSnackbars = new Set();
+
+  function removeSnackbar(key) {
+    registeredSnackbars.delete(key);
+    registeredSnackbars = registeredSnackbars;
+  }
+
+  function showSnackbar(options) {
+    const {
+      component = Snackbar,
+      props = {},
+      duration = 3000,
+    } = options;
+
+    const key = { component, props };
+    key.timeoutID = setTimeout(removeSnackbar, duration, key);
+    registeredSnackbars.add(key);
+    registeredSnackbars = registeredSnackbars;
+
+    return function close() {
+      clearTimeout(key.timeoutID);
+      registeredSnackbars.delete(key);
+      registeredSnackbars = registeredSnackbars;
+    };
+  }
+
+  setContext(snackbarContextKey, showSnackbar);
+</script>
+
+<slot {showSnackbar} />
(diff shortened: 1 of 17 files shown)


src/snackbar/snackbar-container.svelte
```diff
@@ -0,0 +1,46 @@
...
+
+    const key = { component, props };
+    key.timeoutID = setTimeout(removeSnackbar, duration, key);
+    registeredSnackbars.add(key);
+    registeredSnackbars = registeredSnackbars;
+
+    return function close() {
+      clearTimeout(key.timeoutID);
+      registeredSnackbars.delete(key);
+      registeredSnackbars = registeredSnackbars;
```
aabounegm, line 29: Isn't this exactly `removeSnackbar(key)`? DRY

illright:
What's the problem with those containers? They are transparent. If we were to go with something like `<v-app>`, the logic inside would be hopelessly complicated. And this way, you get to choose what do you want – whether it's modals or snackbars. You also have a nice opportunity to specify the position of the snackbars.

You see, the stuff like `v-app` implies a very strong structure to the way you use the framework, and I don't really like that. In this way, Attractions is more like a UI kit, while Vuetify is a full-blown UI framework with a philosophy like "okay, I know better, so just put this here and that there and everything will come up magically"

src/snackbar/snackbar-positions.js
```diff
@@ -0,0 +1,10 @@
+export default Object.freeze({
+  TOP_LEFT: 'top left',
+  TOP_MIDDLE: 'top h-middle',
+  TOP_RIGHT: 'top right',
+  MIDDLE_RIGHT: 'v-middle right',
+  BOTTOM_RIGHT: 'bottom right',
+  BOTTOM_MIDDLE: 'bottom h-middle',
+  BOTTOM_LEFT: 'bottom left',
+  MIDDLE_LEFT: 'v-middle left',
```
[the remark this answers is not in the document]
illright: You mean you'd prefer `LEFT_MIDDLE`?
I wrote those in some sort of natural order, as you would humanly describe those positions, not for consistency between the values

src/snackbar/snackbar.svelte
```diff
@@ -0,0 +1,31 @@
...
+
+<div
+  class={classes('snackbar', _class)}
+  transition:dynamic={{ transition, options: transitionOptions }}
+>
+  <span class={classes('text', textClass)}>{text}</span>
+  {#if action != null}
+    <Button on:click={action.callback} class={buttonClass}>
+      {action.text}
+    </Button>
```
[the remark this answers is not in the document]
illright: You have no way to catch it, the snackbar is created internally. One could forward the events to the snackbar container, but that kinda kills the nice separation of logic

src/snackbar/snackbar-container.svelte
illright: It's also clearing the timeout and closure-ing the key. There has to be a new function object anyway, I see no reason to replace two lines for one in some weird way

src/snackbar/snackbar-positions.js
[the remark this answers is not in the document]
aabounegm: I meant placing line 9 before line 6, but it's not a big deal

src/snackbar/snackbar-container.svelte
aabounegm: I meant just these 2 lines, not the `close` function. The reason to use `removeSnackbar` inside it would be to avoid repetition and changing the same logic in only one place

src/snackbar/snackbar-positions.js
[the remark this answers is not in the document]
illright: Again, simply clockwise enumeration

src/snackbar/snackbar-container.svelte
illright: I know, but these two lines really don't seem worth wrapping in a function call to me

aabounegm: It's not wrapping, they are already in a function. It might not be worth it now, but future refactoring will be easier if we follow the DRY principle

illright: I still don't think it's worth it, but I've run out of energy for this issue, so whatever

PR #27, opened 2020-05-21
Star rating

illright:


src/_utils/range.js
```diff
@@ -0,0 +1,20 @@
+/* Generates a semi-open range. */
+export default function* range(start, end, step = 1) {
+  if (end == null) {
+    end = start;
+    start = 0;
+  }
+
+  if (step === 0) {
+    console.error('Range must have a non-zero step.');
```
aabounegm, line 9: Perhaps a `throw` would be more appropriate in a function?

src/_utils/range.js
```diff
@@ -0,0 +1,20 @@
...
+    return;
+  }
+
+  if ((start >= end && step > 0) || (start <= end && step < 0)) {
+    return;
+  }
+
+  for (let i = start; i !== end; i += step) {
```
aabounegm, line 17: If `end - start` is not exactly divisible by `end`, that would go on indefinitely.

src/star-rating/star-rating.svelte
```diff
@@ -0,0 +1,63 @@
...
+  const ARROW_LEFT_CODE = 37;
+  const ARROW_UP_CODE = 38;
+  const ARROW_RIGHT_CODE = 39;
+  const ARROW_DOWN_CODE = 40;
+
+  function reverseArrowKeys(event) {
+    switch (event.keyCode) {
+
+    case ARROW_LEFT_CODE:
+    case ARROW_UP_CODE:
```
aabounegm, line 27: On what Earth does the up arrow correspond to "less"?

src/star-rating/star-rating.svelte
```diff
@@ -0,0 +1,63 @@
...
+  export let name;
+  export let events = null;
+
+  const ARROW_LEFT_CODE = 37;
+  const ARROW_UP_CODE = 38;
+  const ARROW_RIGHT_CODE = 39;
+  const ARROW_DOWN_CODE = 40;
+
+  function reverseArrowKeys(event) {
+    switch (event.keyCode) {
```
aabounegm, line 24: `event.keyCode` has been deprecated. Better use [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) or [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key). That would have the added benefit of using the string representation right away instead of the ugly `const`s

src/star-rating/star-rating.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+    case ARROW_LEFT_CODE:
+    case ARROW_UP_CODE:
+      event.preventDefault();
+      value = (value + (max + 1) - 1) % (max + 1);
```
aabounegm, line 29: Why `+1` followed by `-1`? This is simply `value + max`

src/star-rating/star-rating.svelte
```diff
@@ -0,0 +1,63 @@
...
+</script>
+
+<div role="group" class={classes('star-rating', _class)}>
+  {#each [...range(max, 0, -1)] as i}
+    <input
+      bind:group={value}
+      value={i}
+      type="radio"
+      name={name}
+      id={`${name}-${i}`}
```
aabounegm, line 51: I would prefer putting the `input` inside its `label` instead of assigning IDs the user never asked for and might be already in use

src/_utils/range.js
illright: Not here :) idk, this is something so utilitary, I really don't see any need for exceptions

src/star-rating/star-rating.svelte
illright: Hm, I actually didn't think about that. The default navigation through radio buttons is what you see here, but I didn't stop to think if it makes sense)

src/star-rating/star-rating.svelte
illright: For the formula to make more sense. It isn't evidently clear why `(value + max) % (max + 1)` yields a previous value

src/_utils/range.js
aabounegm: Who said utility functions can't throw exceptions? If you can be absolutely sure it will never be called with `step = 0`, then the check is useless. Otherwise, exceptions seem like the appropriate error-reporting technique, and not the console

src/star-rating/star-rating.svelte
illright: I would prefer that too, but the CSS dictates us this structure, so we must comply

illright:
As for the range, I'd prefer a generator

PR #28, opened 2020-05-21
Dialogs

illright:


illright:
The modal is a container, the dialog is the filling (so to speak)

PR #31, opened 2020-05-22
Popovers

illright:


src/popover/popover.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    &.right {
+      bottom: 50%;
+      right: 0;
+      transform: translate(100%, 50%);
+    }
+  }
+
+  &:hover > .popover {
```
aabounegm, line 33: Tooltips should be shown on hover, but popovers should be shown on click, according to https://ux.stackexchange.com/questions/88844/when-should-i-use-a-popover-vs-a-tooltip
This is backed by both [Bootstrap](https://getbootstrap.com/docs/4.0/components/popovers/#example) and [Material](https://material-ui.com/components/popover/). To show on hover can be an option that requires a prop

src/popover/popover.svelte
```diff
@@ -0,0 +1,18 @@
+<script>
+  import classes from '../_utils/classes.js';
+
+  let _class = null;
+  export { _class as class };
+  export let popoverClass = null;
+
+  export let position;
```
aabounegm, line 8: Why not set default to bottom? It is what most people will probably use

illright: Well, in innopoints we only used the top popovers. Setting defaults makes sense when it's reasonable, in other words, when one option is clearly more attractive than the others, which isn't the case here. Best to make the devs be explicit about their intentions with this component

src/popover/popover.scss
illright: I would disagree here. Again, because in innopoints we attached the popovers to links, which had a click action of their own. I don't see any strong reason why exactly popovers should only be shown on click, so #wontfix

src/popover/popover.svelte
aabounegm: In that case, specifying the position directly without having to import the enum will be much easier (`position="top"` instead of `position={PopoverPositions.TOP}`) since nothing forces users to use the enum anyways

illright: > In that case

In which case?

Yeah, I agree that specifying positions as strings would be easier, but I'd say that this is discouraged since strings may have typos and with enums we are free to change the internal values without breaking anyone's code

aabounegm: > In which case?

The case of not having defaults.
Can you think of any good reason to change these values? I see it as just overhead for the user that they will have to import two distinct objects from different locations (the component and its enum)

illright: Changing values is not the reason. Error protection is. I'd say an enum is always better in such cases, even with another import

PR #35, opened 2020-05-24
Form field

illright:
Forgot to open a PR


src/text-field/text-field.svelte
```diff
@@ -44,7 +48,7 @@
   {#if multiline}
     <textarea
       {id}
-      bind:value={value}
+      bind:value={internalValue}
       class={inputClass}
       bind:this={inputElement}
       on:input={(e) => dispatch('input', { value: e.target.value, nativeEvent: e })}
```
aabounegm, line 54: Shouldn't this take the `value` instead of `e.target.value` for the casting to number to be effective?

src/form-field/form-field.svelte
```diff
@@ -0,0 +1,49 @@
...
+    {#if errors != null}
+      {#each errors as error}
+        {#if error}<div class="message error">{error}</div>{/if}
+      {/each}
+    {/if}
+    {#if required}
+      <div class="message info">* Required</div>
+    {:else if optional}
+      <div class="message info">Optional</div>
+    {/if}
```
aabounegm, line 44: Use slots with default for easier customization and i18n

src/form-field/form-field.svelte
```diff
@@ -0,0 +1,49 @@
...
+      {#if help != null}
+        <div class={classes('help', helpClass)}>{help}</div>
+      {/if}
+    </div>
+  </slot>
+  <div class="field">
+    <slot />
+    {#if errors != null}
+      {#each errors as error}
+        {#if error}<div class="message error">{error}</div>{/if}
```
aabounegm, line 37: Maybe also a slot with the error as a parameter?

illright: In this case it would make matters only less clear, since the default slot is occupied. With the way of specifying errors as in the demo, this way ends up pretty clean.

src/text-field/text-field.svelte
illright: Yep, that's right

src/form-field/form-field.svelte
illright: Tried putting something together about this, realized that I don't like any of it. Custom messages can be passed using a slot below, and as for i18n, we can simply wrap the text in a i18n mapper later on in the library. Don't feel like having those slots here.

illright:
Yeah, `FormField` without `Form`. There isn't really much use in `Form`, as this is more of a style and layout component as opposed to a form logic component. We'll leave that to form libraries of Svelte (with which we will hopefully be compatible)

PR #36, opened 2020-05-27
Simplify the Modal implementation

illright:
Yeah, you were right. Was able to remove the wrapper component


illright:
Was thinking of a similar way to simplify the Snackbar, but nothing comes to mind

PR #37, opened 2020-05-27
Time pickers

illright:
Decisions:

* no `range` version as the picker dropdown doesn't really get any more useful for the range, the users will simply put up two of those time pickers if they need a range
* no combined datetime picker, because, again, the dropdown doesn't benefit from this combination

### DIFF
--- a/src/text-field/text-field.svelte
+++ b/src/text-field/text-field.svelte
@@ -31,7 +31,7 @@
   onMount(() => autofocus && inputElement.focus());
 
   function handleInput(event) {
-    value = $$restProps.type === 'number' ? +event.target.value : event.target.value;
+    value = $$restProps.type === 'number' ? parseFloat(event.target.value) : event.target.value;
     dispatch('input', { value, nativeEvent: event });
   }
 
--- a/src/text-field/text-field.svelte
+++ b/src/text-field/text-field.svelte
@@ -24,15 +24,16 @@
     console.error('Labels are only available for outlined text fields');
   }
 
-  let internalValue = null;
   export let value = null;
-  $: value = (
-    $$restProps.type === 'number' && internalValue != null ? +internalValue : internalValue
-  );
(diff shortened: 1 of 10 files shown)


src/text-field/text-field.svelte
```diff
@@ -24,16 +24,17 @@
...
-  $: value = (
-    $$restProps.type === 'number' && internalValue != null ? +internalValue : internalValue
-  );
   export let events = null;
 
   let inputElement;
   onMount(() => autofocus && inputElement.focus());
 
+  function handleInput(event) {
+    value = $$restProps.type === 'number' ? +event.target.value : event.target.value;
```
aabounegm, line 34: I wanted to suggest using `parseInt` instead of the unary plus because it seem more fitting, depending on the first 2 answers on [this SO post](https://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which) (especially the summary table in the 2nd answer), but then I decided to check the internal svelte implementation for `<input type="number">`, and turns out that they just do like this:
```js
function to_number(value) {
    return value === '' ? undefined : +value;
}
```
(It is worth mentioning that this can be imported using `import { to_number } from 'svelte/internal';`)
What do you think about this?

illright:
> 00 doesn't really make sense in 12-hours mode. The range should be [1, 12] not [0, 11] in that case (where 12 a.m. corresponds to 00, and 12 p.m. corresponds to 12)

\*disagrees in 24-format\*

idk, makes sense to me. And in fact it makes more sense to have the range be [0, 11]. However, a lot of the AM/PM stuff doesn't make sense to me, so if you could help out here, that'd be great

src/text-field/text-field.svelte
illright: Not keen on importing anything from `svelte/internal`. As for `parseInt`, yes, it would work better here.

PR #39, opened 2020-05-28
Npm release

aabounegm:


.github/workflows/deploy.yml
```diff
@@ -1,11 +1,12 @@
-name: Deploy demo to GH Pages
+name: Release
 on:
   release:
     types: [published]
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    name: Deply demo to Pages
```
illright, line 9: Depl**o**y

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "attractions",
-  "version": "0.0.0",
+  "version": "1.0.0-alpha",
```
illright, line 3: Is the `-alpha` prefix semver compliant?

aabounegm: That's a suffix, and [yes](https://semver.org/)

PR #42, opened 2020-06-08
Give information about the nature of the snackbar disappearing

illright:
This will return a promise that resolves when the snackbar is closed. From that one can deduce if the snackbar expired or was closed manually (perhaps with the snackbar action). Useful for situations like undo actions

### DIFF
--- a/src/snackbar/snackbar-container.svelte
+++ b/src/snackbar/snackbar-container.svelte
@@ -8,7 +8,7 @@
 
   function removeSnackbar(key, closedEarly) {
     registeredSnackbars.delete(key);
-    if (key.resolveExpiredPromise != null) {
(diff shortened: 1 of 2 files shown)


src/snackbar/snackbar-container.svelte
```diff
@@ -6,8 +6,11 @@
   export let position;
   let registeredSnackbars = new Set();
 
-  function removeSnackbar(key) {
+  function removeSnackbar(key, closedEarly) {
     registeredSnackbars.delete(key);
+    if (key.resolveExpiredPromise != null) {
```
aabounegm, line 11: I found that when checking for the "nullness" of a function, it is more common to do it as such
```suggestion
    if (typeof key.resolveExpiredPromise === 'function') {
```
to avoid more possible errors at once

PR #44, opened 2020-06-08
Balance z-indices to avoid the ripple overlaying modals

illright:


src/_utils/ripple.js
```diff
@@ -3,7 +3,7 @@ export default function ripple(node, options = {}) {
   let props = {
     event: options.event || 'click',
     transition: options.transition || 150,
-    zIndex: options.zIndex || '9999',
+    zIndex: options.zIndex || '100',
```
aabounegm, line 6: Why is it a string and not a number?

illright: uh, just because it was written that way, no particular reason

illright:
Multiples of a 100 were used to give space in between, to whoever might need it

PR #48, opened 2020-06-12
Add postinstall script to move files from src to .

aabounegm:
Instead of having to use `'attractions/src/*'` when importing something specific, everything in the `src` folder will now be moved up one level to the current (installation) directory.


aabounegm:
huh, that's weird. didn't expect the postinstall script to run after installing the dependencies, just the package itself

aabounegm:
Will restart from a clean branch

PR #49, opened 2020-06-15
Remove src folder

aabounegm:
To be able to import components as `import { ... } from 'attractions/{{component}}';` directly with having `/src/` in the middle, and to achieve consistency in the paths between using `npm link` and the published package, the components (and their package.json) have been moved from the `src` directory to the `attractions` directory, where this directory is now the root of the published module


illright:
Wait, before you merge, what's up with the demo? It doesn't load for me for some reason

PR #51, opened 2020-06-17
Make the Autocomplete emit change events

illright:


attractions/autocomplete/autocomplete.svelte
```diff
@@ -39,6 +43,7 @@
     {maxOptions}
     {optionComponent}
     {...$$restProps}
+    on:change
```
aabounegm, line 46: You keep forgetting about the rename before forwarding)

illright: If by _rename_ you mean wrapping the value into an object, then yes, you are right, fixed it

aabounegm: I meant renaming the event to `nativeEvent`

illright: That one isn't native, it comes from the `AutocompleteField`

aabounegm: oh, GitHub didn't show all the lines so I didn't notice 😅

PR #52, opened 2020-06-17
Fix the value handling for TextFields of type="number"

illright:
This one is needed because previously empty `<TextField type="number">` elements would send a warning to the console saying that it cannot parse NaN, which is a result of some messy handling.

### DIFF
--- a/attractions/text-field/text-field.svelte
+++ b/attractions/text-field/text-field.svelte
@@ -30,8 +30,22 @@
   let inputElement;
   onMount(() => autofocus && inputElement.focus());
 
+  function toNumber(string) {
+    if (string === '') {
+      return null;
+    }
+    return parseFloat(string);
+  }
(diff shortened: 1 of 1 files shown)


attractions/text-field/text-field.svelte
```diff
@@ -30,8 +30,22 @@
   let inputElement;
   onMount(() => autofocus && inputElement.focus());
 
+  function toNumber(string) {
+    if (string === '') {
```
aabounegm, line 34: How about `string.trim()`, because a string with just whitespaces would also return `NaN`?

illright: Not that it's much of a burden to add, but whitespace simply cannot make its way into an `<input type="number">`, this check was particularly to handle the case when all characters are erased from the field

PR #53, opened 2020-06-17
Allow restyling the tabs vertically

illright:
Just internal style refactoring


geoidesic:
Demo doesn't work: 404

illright:
@geoidesic this demo is long gone, it was just an internal development thing :) if you need assistance restyling the tabs, you can create an issue or, better yet, start a thread in Discussions

PR #56, opened 2020-06-24
Allow omitting the callback in the snackbar action

illright:
In some cases one only wants the snackbar action to close it (in order to force some action, as in `undo`, for example, would _actually_ do the action after the snackbar expires).

### DIFF
--- a/attractions/snackbar/snackbar.svelte
+++ b/attractions/snackbar/snackbar.svelte
@@ -27,7 +27,10 @@
     <Button
       class={buttonClass}
       on:click={() => {
-        action.callback();
+        if (typeof action.callback === 'function') {
+          action.callback();
+        }
+
         if (closeOnAction) {
           closeCallback();
         }


attractions/snackbar/snackbar.svelte
```diff
@@ -27,7 +27,10 @@
     <Button
       class={buttonClass}
       on:click={() => {
-        action.callback();
+        if (typeof action.callback === 'function') {
+          action.callback();
+        }
+
         if (closeOnAction) {
```
aabounegm, line 34: Why isn't the check here similar to the previous one? It would help protect against bad input like `true`.
```suggestion
        if (typeof closeOnAction === 'function') {
```

illright: Um, because `true` is, in fact, the correct input here)

aabounegm: Oh sh*t. I misread the following line to also be `closeOnAction` 😅

PR #57, opened 2020-06-24
Implement the Divider component

illright:


aabounegm:
No demo? :(

illright:
Yeah, just couldn't bother. This is a very tiny component that I again implemented in the previous non-existent version of Attractions and forgot to transfer, so just not worth it

PR #58, opened 2020-06-26
Allow passing a function to be called before file acceptance

illright:
UPD: we're talking about the FileDropzone component here

In the case of Innopoints, this could mean starting the upload so that a Promise with the ID exists. It also supports exceptions, so if a file is invalid for some reason (filename or size), an exception can be thrown in that function and this file will not be added.

### DIFF
--- a/attractions/file-input/file-dropzone.svelte
+++ b/attractions/file-input/file-dropzone.svelte
@@ -13,6 +13,7 @@
   export let files = [];
   export let fileComponent = FileTile;
   export let accept = null;
+  export let beforeChange = null;
   export let disabled = false;
 
   let dragActive = false;
@@ -26,12 +27,19 @@
     wrongType = [...e.dataTransfer.items].some(file => !accepted(accept, file));
   }
 
-  function acceptUpload(e) {
+  async function acceptUpload(e) {
     for (let file of (e.dataTransfer || e.target).files) {
-      if (accepted(accept, file)) {
-        files.push(file);
-      }
+      try {
+        if (typeof beforeChange === 'function') {
+          await beforeChange(file);
+        }
+
+        if (accepted(accept, file)) {
+          files.push(file);
+        }
+      } catch (e) {}  // eslint-disable-line no-empty
     }
+
     files = files;
     setTimeout(() => wrongType = false, 1000);
     dispatch('change', { files, nativeEvent: e });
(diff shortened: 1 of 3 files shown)


attractions/file-input/file-dropzone.svelte
```diff
@@ -13,6 +13,7 @@
   export let files = [];
   export let fileComponent = FileTile;
   export let accept = null;
+  export let beforeChange = null;
```
aabounegm, line 16: To avoid the sanity check in the `acceptUpload` function, it might be better to use a no-op function as the default
```suggestion
  export let beforeChange = async () => {};
```

attractions/file-input/file-dropzone.svelte
```diff
@@ -26,20 +27,27 @@
     wrongType = [...e.dataTransfer.items].some(file => !accepted(accept, file));
   }
 
-  function acceptUpload(e) {
+  async function acceptUpload(e) {
     for (let file of (e.dataTransfer || e.target).files) {
```
aabounegm, line 31: Consider using `.forEach` so that it doesn't wait for `beforeChange` to resolve (or reject) before moving on to the next file

attractions/file-input/file-dropzone.svelte
```diff
@@ -26,20 +27,27 @@
...
-      }
+      try {
+        if (typeof beforeChange === 'function') {
+          await beforeChange(file);
+        }
+
+        if (accepted(accept, file)) {
+          files.push(file);
+        }
+      } catch (e) {}  // eslint-disable-line no-empty
```
aabounegm, line 40: Wouldn't it be a little neater if this try-catch was moved to surround just the `beforeChange` call, with the `catch` block having just `continue`?

attractions/file-input/file-dropzone.svelte
illright: Nah

attractions/file-input/file-dropzone.svelte
```diff
@@ -28,17 +28,20 @@
   }
 
   async function acceptUpload(e) {
-    for (let file of (e.dataTransfer || e.target).files) {
+    const incomingFiles = Array.from((e.dataTransfer || e.target).files);
+    await Promise.all(incomingFiles.map(async (file) => {
```
aabounegm, line 32: `Promise.allSettled` would allow us to remove the try-catch altogether, but its support is much more limited than `Promise.all`, so I'm just noting that we can consider it if we will add Babel or something later on

PR #61, opened 2020-06-29
Feature/pagination

aabounegm:
Any suggestions about what can be turned into a variable?


aabounegm:
I'm also thinking there should be a prop that controls whether it should be hidden if only one page exists, but I'm not sure how to name it

illright:
How about `showOnlyPage`?

aabounegm:
Sounds a bit weird. It sort of suggests that there are other things to show, but we only want to show "the page")

illright:
Sounds a little overly poetic for a prop name, but works too

PR #71, opened 2020-07-02
Home page of the docs

illright:
So I built the home page, nothing more yet. Thinking about how do we set up CI now. Thoughts?


docs/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/runtime": "^7.0.0",
+    "@rollup/plugin-alias": "^3.1.1",
+    "@rollup/plugin-babel": "^5.0.0",
+    "@rollup/plugin-commonjs": "^12.0.0",
+    "@rollup/plugin-node-resolve": "^8.0.0",
+    "@rollup/plugin-replace": "^2.2.0",
+    "attractions": "^2.0.1-alpha",
```
aabounegm, line 28: Will it not be linked directly to it through `../attractions`? It's not like `docs` is a package that will get released to npm, this package.json is there just for the dependencies, right?

docs/rollup.config.js
```diff
@@ -0,0 +1,133 @@
...
+    ],
+    external: Object.keys(pkg.dependencies).concat(
+      require('module').builtinModules || Object.keys(process.binding('natives')),
+    ),
+
+    preserveEntrySignatures: 'strict',
+    onwarn,
+  },
+
+  serviceworker: {
```
aabounegm, line 117: You think the docs really deserve a service worker? Do you want it to work offline or become a PWA or what?

docs/rollup.config.js
```diff
@@ -0,0 +1,133 @@
+import resolve from '@rollup/plugin-node-resolve';
```
aabounegm, line 1: Why do the docs have a separate rollup config file? The demo was added to the same config as the core library. Since the docs are meant to replace this demo, either it follows the same structure or we move the library's config to its folder (attractions)

docs/src/routes/index.svelte
```diff
@@ -0,0 +1,43 @@
+<script>
+  import { Card, Button, H1 } from 'attractions';
+  import { BookOpenIcon, GithubIcon } from 'svelte-feather-icons';
+  import InfoTiles from 'src/containers/home/info-tiles.svelte';
+</script>
+
```
aabounegm, line 6: Let's also add a `<svelte:head>` with all the OpenGraph meta for link previews?

docs/package.json
illright: I thought so as well, but upon importing from `../../attractions`, Sapper complains that the imported things are not SSR components, so I decided to install it as a dependency.

As I was writing this, I realized that we don't really need SSR, since we will be exporting the website statically anyway. In that case we should try to use the existing version (we can even alias `attractions` to `~/attractions` for convenience). If the exporting works, we can simply turn off SSR

docs/rollup.config.js
illright: Uh, I figured it doesn't hurt. It was there to begin with, might as well make use of the work that the Sapper devs have made to make the web more progressive.
Besides, caching docs is quite handy.

docs/src/routes/index.svelte
illright: Yeah, that's all in the coming

docs/rollup.config.js
illright: A separate config file is handier somehow. As for the demo, yeah, maybe we should move the demo's config into the demo folder, and the core library doesn't even use Rollup, right?

aabounegm: The core library does use Rollup for building before publishing to npm (to have the non-Svelte versions).
I thought we will remove the demo entirely once the docs are ready?

docs/src/containers/home/info-tiles.svelte
```diff
@@ -0,0 +1,28 @@
+<script>
+  import * as attractions from 'attractions';
+  import { s } from 'attractions/utils';
+  import { GridIcon, Edit2Icon, FeatherIcon } from 'svelte-feather-icons';
+  import InfoTile from 'src/components/home/info-tile.svelte';
+
+  const totalComponents = [...Object.keys(attractions)].length;
```
aabounegm, line 7: Why the spread and array reconstruction?
```suggestion
  const totalComponents = Object.keys(attractions).length;
```
should suffice

docs/src/routes/index.svelte
aabounegm: This PR or in the future?

docs/package.json
aabounegm: What kind of complaints, and why are they not SSR components?

illright: It's just a thing in Sapper, if you want to use external components with SSR, you must install libraries to dev dependencies

docs/rollup.config.js
illright: Yes, but that time hasn't come yet. The source code of the demo is still useful

docs/src/routes/index.svelte
illright: Future, this one's ready for review

illright:
> How about putting the scss of each component next to it in the same folder to avoid the bunch of ../../?

Not very keen on this, this doesn't really scale well. I think the approach we've taken with Innopoints is a lot better for Sapper-based websites. As for `../../`, I just can't get around to contacting the developer of `svelte-preprocess` about the aliasing issue, but once that's done, this won't be a problem.

> One comment though: doesn't the blue color suggest clickable links?
You mean the `<mark>` element on the "A _pretty cool_ UI kit"? It does, but not as much as an underline, and that's not present, so I see no fear of confusion here

docs/package.json
aabounegm: I'll let you deal with that, but I believe we need to link directly instead of use it as a dependency because when editing a component and its docs together, the docs will not see the update in the component

PR #73, opened 2020-07-03
Set default working directory for the jobs

aabounegm:


illright:
@github-actions[bot] What is up with styles?

PR #74, opened 2020-07-05
Allow disabling the clickaway behaviour of the modal overlay

illright:
I deliberately opted for a certain amount of code duplication to prevent from messing around with the code structure + remove an event listener when it's not used

### DIFF
--- a/attractions/modal/modal.svelte
+++ b/attractions/modal/modal.svelte
@@ -5,6 +5,7 @@
   let _class = null;
   export { _class as class };
   export let open = false;
+  export let clickaway = true;
   $: dispatch('change', { value: open });
 
   function close() {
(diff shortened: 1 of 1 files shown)


attractions/modal/modal.svelte
```diff
@@ -5,6 +5,7 @@
   let _class = null;
   export { _class as class };
   export let open = false;
+  export let clickaway = true;
```
aabounegm, line 8: Name is not really clear. How about `persistent` (with inverted behavior) or `closeOnClickaway`?

illright: `persistent` sounds even less clear, and `closeOnClickaway` is too long and reminiscent of an event. The clickaway term is somewhat known + we have docs to clarify the meaning, so I don't think a change is worth it

PR #78, opened 2020-07-08
Set up pre-commit linting hooks

illright:
This still will not add workspaces. Also I hope workspaces aren't exclusive to Yarn 2. We are using Yarn 1, and I tried to move to Yarn 2, but it was such a mess, Yarn 1 seemed a lot friendlier than whatever they turned it into, so I don't want to make the switch.


.eslintignore
```diff
@@ -1,5 +1,6 @@
-/node_modules/
+/node_modules
```
aabounegm, line 1: How about any `node_modules` in general? (Remove the slash)

.stylelintignore
```diff
@@ -1,5 +1,8 @@
 node_modules/
 demo/public/
+demo/node_modules/
```
aabounegm, line 3: Same here. Just `node_modules` and `__sapper__`. I don't think there is ever a case where we might need to include one of them

aabounegm:
Workspaces are available since Yarn v1.0. Do you want me to add them here or in a different PR (if we decide to use them)?

illright:
Preferably a different PR

PR #81, opened 2020-07-11
Add the breadcrumbs to the docs header

illright:


illright:
Accidentally pushed the Prism CSS file, that one was meant for a different PR

docs/src/components/docs/breadcrumbs.svelte
```diff
@@ -0,0 +1,22 @@
...
+  import { HomeIcon } from 'svelte-feather-icons';
+
+  export let segment;
+
+  function capitalize(word) {
+    return word[0].toUpperCase() + word.slice(1);
+  }
+</script>
+
+<nav class="breadcrumbs">
```
aabounegm, line 12: Wouldn't this make for a good component of its own? Perhaps with an array of segments for a prop

illright: Maybe, but it would need pretty serious reconstruction, and as it currently stands, it's not meant for general usage

docs/static/css/components/docs/breadcrumbs.scss
```diff
@@ -0,0 +1,17 @@
+@import "_attractions-theme.scss";
+
+nav {
+  display: none;
+
+  @media only screen and (min-width: 1024px) {
```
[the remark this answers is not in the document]
illright: Again, because it's not meant for general usage, not worth it

PR #82, opened 2020-07-12
Document the Button component

illright:
Also add the installation documentation


attractions/button/button.svelte
```diff
@@ -43,8 +44,7 @@
 {#if href}
   <a
     {href}
```
aabounegm, line 46: Shouldn't the `href` be removed if `disabled == true`?

docs/src/mdsvex/h1.svelte
```diff
@@ -0,0 +1,23 @@
...
+
+    href = window.location.origin + window.location.pathname + href;
+  });
+</script>
+
+<h1 {id}>
+  <slot />
+  {#if href != null}
+    <a {href} class="ml">#</a>
+  {/if}
```
aabounegm, line 20: Don't you think it will look better on the left?

docs/src/mdsvex/h2.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
```
aabounegm, line 1: I hate code duplication. At least the script for h1 - h3 is the same. I recommend moving it to a separate .js file and including it using `<script src="./h.js">` (or similar) instead

docs/src/mdsvex/li.svelte
```diff
@@ -0,0 +1,3 @@
+<li><slot /></li>
```
aabounegm, line 1: If the only purpose for all of these components is to include the styles, wouldn't it be easier to just make the styles global? (especially that these components will only affect the mdsvex routes)

docs/src/routes/docs/components/button.svx
```diff
@@ -0,0 +1,127 @@
+---
+shadow1: '0 2px 5px rgba(0, 0, 0, .14), 0 1px 5px rgba(0, 0, 0, .12);'
+shadowRaised: '0 2px 4px -1px rgba(0, 0, 0, .2), 0 3px 5px 0 rgba(0, 0, 0, .14), 0 1px 10px 0 rgba(0, 0, 0, .12);'
+---
+
+<script>
+  import { Button, Label } from 'attractions';
+  import { CheckIcon, XIcon } from 'svelte-feather-icons';
+  import H2 from 'src/mdsvex/h2.svelte';
```
aabounegm, line 9: Why do you need to import it here? shouldn't mdsvex automatically replace any `h2` with this `H2`?

docs/src/routes/docs/components/button.svx
```diff
@@ -0,0 +1,127 @@
...
+  <div slot="source">
+
+```svelte
+<Button filled>button</Button>
+<Button outline>button</Button>
+<Button>button</Button>
+
+<Button filled disabled>button</Button>
+<Button outline disabled>button</Button>
+<Button disabled>button</Button>
```
aabounegm, line 51: It might be the OCD talking, but it would look a lot nicer if the same attributes were lined up together)
```suggestion
<Button disabled filled>button</Button>
<Button disabled outline>button</Button>
<Button disabled>button</Button>
```

docs/src/routes/docs/components/button.svx
```diff
@@ -0,0 +1,127 @@
...
+
+<Button filled danger>button</Button>
+<Button outline danger>button</Button>
+<Button danger noRipple>button</Button>
```
aabounegm, line 55: Same
```suggestion
<Button danger filled>button</Button>
<Button danger outline>button</Button>
<Button danger noRipple>button</Button>
```

docs/src/routes/docs/components/button.svx
```diff
@@ -0,0 +1,127 @@
...
+
+  </div>
+</Showcase>
+
+
+<H2 id="properties">Properties</H2>
+
+### Style Properties
+
+| <Label>Name</Label> | <Label>Type</Label> | <Label>Description</Label> |
```
aabounegm, line 73: I have a strong feeling that this table can (should) be made into a component and just pass an array of data items

docs/src/routes/docs/components/button.svx
```diff
@@ -0,0 +1,127 @@
...
+| --- | --- | --- |
+| **`disabled`** | `boolean` | Disables the button, styling it appropriately and disabling click events. <br /> Has no effect on link buttons (those with an ==`href`== prop). |
+| **`href`** | `string` | Turns the button into a link (prefetch-enabled for Sapper). |
+| **`noPrefetch`** | `string` | Disables Sapper prefetching for link buttons (those with an ==`href`== prop). |
+| **`events`** | `Array` | List of handlers for the [Events action](): `[{ name, handler }]`. |
+| **`...`** | `*` | Everything else will be passed to the underlying `<button>` or `<a>` element. |
+
+### Class Properties
+
+All `<Button>` components use the `.btn` class.
```
aabounegm, line 97: Is this really something that should be documented? It's better to add their own classes if they want to override anything. Due to scoping, our selectors will have higher precedence over just `.btn {}` anyways, which can make it more frustrating when someone tries to modify it and can lead to lots of `!important`

docs/src/routes/docs/installation.svx
```diff
@@ -0,0 +1,34 @@
...
+## Theming
+
+You may configure style parameters such as colors, fonts and shadows by overriding respective variables in your `_attractions-theme.scss` file:
+
+```scss
+$main: #b17200;
+$font: 'Noto Sans', sans-serif;
+```
+
+Refer to the documentation for each component for the variables available for overriding.
```
aabounegm, line 34: Too many "for"s in one sentence 😜
```suggestion
Refer to the documentation of each component for the variables available for overriding.
```

attractions/button/button.svelte
illright: Well...
As it currently stands, not really. The docs state that `disabled` has no effect on link buttons. We could alter that behaviour if we wanted. As of now, we simply pass through the `disabled` prop to allow the user to style disabled links if they so desire

In other words, it's a major gray area, there's a lot of ways to go about it, we need to settle on one

P.S. Removing the `href` is a bad idea because it removes the default link styling.

docs/src/mdsvex/h1.svelte
illright: Nah. It might have been better on the left if that left was absolutely positioned (i.e. not shift the heading to the right) + appeared only on hover, but as of now, I think this works nicely too

docs/src/mdsvex/h2.svelte
illright: Aaaaand what exactly do you expect this file to contain?
I feel your pain, at some point I even wanted to set up Rollup to generate those files, but then figured that's even more insane that a little code duplication

docs/src/mdsvex/li.svelte
illright: Now that scoping styles has grown so much on me, definitely no. Worth it all the way

docs/src/routes/docs/components/button.svx
illright: It would replace `##` with a `<H2>`, but in this case we need to pass props, so I use the component explicitly.

docs/src/routes/docs/components/button.svx
illright: I really hope you never get to encounter OCD in your life)
Sure, I can change that

docs/src/routes/docs/components/button.svx
illright: Keep in mind that currently what we have is very expressive + retains full customizability. I'm quite hesitant to give up being able to throw an occasional component here and there. Besides, I see no way in which an array of data items representing a table could be formatted any better than a Markdown table

docs/src/routes/docs/components/button.svx
illright: Coming back to our discussion about the importance of CSS. Yes, I whole-heartedly believe that this should be documented + the classes should have reasonable names (which they do, as of now). As for precedence, true, some might find it frustrating, but keep in mind that properly styling a child component in Svelte involves scoping the style from the outside:
```scss
.something-local :global .btn {}
```
which gets translated into 
```scss
.something-local.svelte-xxx .btn {}
```
which is both nicely isolated and has higher precedence over `.btn.svelte-yyy`

docs/src/routes/docs/installation.svx
illright: True

attractions/button/button.svelte
aabounegm: I saw that the docs represent the current state of the component, but I wasn't sure if that was actually the goal or just an unintended "limitation". I believe the prop should still apply and disable the navigation. This is typically done by just removing the `href` attribute. An alternative can be to set it to `javascript:void(0)`, so it's still there but does nothing

docs/src/mdsvex/h2.svelte
aabounegm: Exactly the same content:
```js
import { onMount } from 'svelte';

export let id = null;

let href = id && `#${id}`;
onMount(function createSapperCompatiblePermalink() {
  if (id == null) {
    return;
  }
  href = window.location.origin + window.location.pathname + href;
});
```

docs/src/mdsvex/li.svelte
aabounegm: Don't forget that this is the docs app, not the core library. If every single `<li>` will get replaced with that one instead, is it really that worth it?

docs/src/mdsvex/h2.svelte
illright: Oh, that's what you mean. Welp, maybe, but it would be a major exception in an established system of file structure, and the benefit of that is quite negligible. Separating the script is especially daunting because the markup uses props that you can't immediately see, which is different from styles, where you don't really use them in the markup or script whatsoever

docs/src/routes/docs/components/button.svx
aabounegm: Another option would be to use `##` and auto-generate the id from the heading text itself, since I don't think there is a heading (at least h1 - h3) that doesn't deserve its fragment

docs/src/mdsvex/li.svelte
illright: This way of styling stuff is good for two reasons: predictability and consistency. If I wanna change the style for `<li>`, I know where exactly to look for it (this is a major game changer, take it from someone who had to pull apart a crap ton of stylesheets in Innopoints that were intertwined in the most obscure of ways). Also it doesn't break the pattern of code structure needlessly. This is a low price to pay for such noticeable benefits

docs/src/routes/docs/components/button.svx
aabounegm: "I believe" is not really an explanation/argument, but it's up to you)

docs/src/routes/docs/components/button.svx
illright: Yep, I wish that was possible, but Svelte currently doesn't allow accessing slots programmatically

PR #85, opened 2020-07-12
Add buttons to copy code

illright:


docs/src/routes/docs/installation.svx
```diff
@@ -2,37 +2,61 @@
...
 # Installation
 
 **Step 1.** Install the library with NPM or Yarn:
 
-```bash
-npm install --save-dev attractions
-# -- or --
-yarn add -D attractions
-```
+<CopyableCode>
```
aabounegm, line 13: There doesn't seem to be a code block that is not `<CopyableCode>`. Does that call for a layout custom component?

illright: Yeah, it's not possible as of now. I've filed an issue, but until then, we'll have to make do

PR #86, opened 2020-07-13
Accordion + Autocomplete docs

illright:


docs/src/routes/docs/components/autocomplete-field.svx
```diff
@@ -0,0 +1,133 @@
...
+
+The message to show in the dropdown when the maximum amount of selected options has been reached.
+
+### `not-enough-input` slot
+
+The message to show in the dropdown when the minimum amount of characters to type for searching hasn't been reached.
+
+### `loading-options` slot
+
+The loading state of the dropdown while the next promise with suggestions is resolving.
```
aabounegm, line 114: How about mentioning the default here? Or is the showcase enough?

docs/src/routes/docs/components/autocomplete-field.svx
```diff
@@ -0,0 +1,133 @@
...
+</Showcase>
+
+
+## Properties {#properties}
+
+### Functional Properties
+
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| ==**`getOptions`**==&nbsp;<span class="required">*</span> | `async function*` | An async generator of suggestions. <br /> Receives the input from the text field and is expected to yield promises that resolve to arrays of objects `{ name, details }`. |
```
aabounegm, line 75: > and is expected to yield promises

They don't necessarily have to be promises, direct values also work with `await` and `Promise.all`. A note like this can be useful for users who have the data already and think they need to wrap it in a `Promise`

docs/src/routes/docs/components/autocomplete-field.svx
```diff
@@ -0,0 +1,133 @@
...
+| **`selection`** | `Array` | The current selection as an array of objects `{ name, details }`. <br /> Can be used to set the selection programmatically. |
+| **`minSearchLength`** | `number` | The minimum amount of characters to call ==`getOptions`==. <br /> 3 by default. |
+| **`maxOptions`** | `number` | The maximum amount of options than can be selected. Set to `null` (the default) for no restrictions. |
```
aabounegm, line 78: How about we change the default to be `Infinity`? I imaging it would be cumbersome for Typescript users to define the type as `number|null` and have to deal with its nullability everywhere, and `Infinity` would be just like a normal number that we don't have to check `{#if maxOptions != null}` for

docs/src/routes/docs/components/autocomplete-field.svx
```diff
@@ -0,0 +1,133 @@
...
+| **`searchQuery`** | `number` | The current value of the text field. Can be used to control the query programmatically. |
+| **`optionComponent`** | Svelte component | The component used to render a suggestion in the list. <br /> Receives two props: the `option` object (`{ name, details }`) and the current `query`. <br /> Expected to dispatch `click` events on selection. <br /> Defaults to [==`AutocompleteOption`==](./docs/components/autocomplete-option). |
+| **`focus`** | `boolean` | Allows to programmatically control the suggestions dropdown. |
```
aabounegm, line 81: How about renaming this to `active`, or at least explain what does it control about the dropdown? The description is not really clear and I had to go check the source code

docs/src/routes/docs/components/autocomplete.svx
```diff
@@ -0,0 +1,104 @@
...
+| **`selection`** | `Array` | The current selection as an array of objects `{ name, details }`. <br /> Can be used to set the selection programmatically. |
+| **`minSearchLength`** | `number` | The minimum amount of characters to call ==`getOptions`==. <br /> 3 by default. |
+| **`maxOptions`** | `number` | The maximum amount of options than can be selected. Set to `null` (the default) for no restrictions. |
+| **`searchQuery`** | `number` | The current value of the text field. Can be used to control the query programmatically. |
+| **`optionComponent`** | Svelte component | The component used to render a suggestion in the list. <br /> Receives two props: the `option` object (`{ name, details }`) and the current `query`. <br /> Expected to dispatch `click` events on selection. <br /> Defaults to [==`AutocompleteOption`==](./docs/components/autocomplete-option). |
+| **`...`** | `any` | Everything else will be passed to the underlying `<AutocompleteField>` component. |
```
aabounegm, line 75: Isn't this note enough instead of listing the props above? Maintaining the same list of props in 2 places is not something pleasant

docs/src/routes/docs/components/autocomplete-field.svx
illright: I think it is

docs/src/routes/docs/components/autocomplete-field.svx
illright: Yeah, but wrapping a value in a `Promise.resolve` isn't that difficult, besides, writing an async function will ensure that. I think it's ok

docs/src/routes/docs/components/autocomplete.svx
illright: Jumping around doc pages when you're trying to figure out a component isn't pleasant either

aabounegm: Then how about refactoring this data to a separate file and importing it in both?

PR #87, opened 2020-07-14
Document the Badge

illright:


docs/src/routes/docs/_layout.svelte
```diff
@@ -34,6 +34,10 @@
           title: 'Button',
           segment: 'components/button',
         },
+        {
+          title: 'Badge',
```
aabounegm, line 38: I thought you were ordering the components alphabetically?

illright: Ah, right, button goes next. Will fix that

PR #88, opened 2020-07-14
Document the Card

illright:


docs/src/routes/docs/_layout.svelte
```diff
@@ -34,6 +34,14 @@
           title: 'Button',
           segment: 'components/button',
         },
+        {
+          title: 'Badge',
+          segment: 'components/badge',
+        },
```
aabounegm, line 40: I'm just curious, how did this object come from the other branch, but not the other files?

illright: I specifically added it to avoid merge conflicts, that would be the only file that could have caused them

illright: But seems like it didn't help

PR #93, opened 2020-07-18
Document the Chip components

illright:


docs/src/routes/docs/components/checkbox.svx
```diff
@@ -65,7 +65,7 @@ An individual checkbox.
 
 | Name | Default | Type | Description |
 | ---- | ------- | ---- | ----------- |
-| **`value`** | `null` | `string` | The value assigned to the checkbox input. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value) for more information. |
+| **`value`** | `null` | `any` | The value assigned to the checkbox input. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value) for more information. |
```
aabounegm, line 68: Why? The documentation specifies that this should be a [`DOMString`](https://developer.mozilla.org/en-US/docs/Web/API/DOMString)

docs/src/routes/docs/components/checkbox-chip.svx
```diff
@@ -0,0 +1,107 @@
...
+
+## Properties {#properties}
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`small`** | `false` | `boolean` | Makes the chip smaller. |
+| **`outline`** | `false` | `boolean` | Adds a border to the chip instead of the background. |
+| **`noPadding`** | `false` | `boolean` | Clears the internal padding. Useful for adding deletion buttons. |
```
aabounegm, line 55: The second sentence is not very clear imo.

docs/src/routes/docs/components/checkbox-chip.svx
```diff
@@ -0,0 +1,107 @@
...
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| ==**`value`**==&nbsp;<span class="required">*</span> |  | `any` | The value of this checkbox. Included in events. |
+| ==**`checked`**== | `false` | `boolean` | Whether the chip is selected or not. |
+| ==**`name`**==&nbsp;<span class="required">*</span> |  | `string` | The name of the group to relate this checkbox chip to. |
```
aabounegm, line 63: Since I don't see any particular ordering (e.g.: alphabetical), how about having all the required props first?

docs/src/routes/docs/components/checkbox-chip.svx
```diff
@@ -0,0 +1,107 @@
...
+| **`disabled`** | `false` | `boolean` | Disables the chip and disallows selection. |
+| **`title`** | `null` | `string` | The tooltip to give to a chip. |
+| **`...`** |  | `any` | Everything else will be passed to the underlying `<input>` element. |
+
+<div class="required">* Required</div>
+
+### Class Properties
+
+All `<CheckboxChip>` components use the `.input-chip.checkbox-chip` class.
```
aabounegm, line 72: These are 2 class**es**.
```suggestion
All `<CheckboxChip>` components use the `.input-chip` and `.checkbox-chip` classes.
```

docs/src/routes/docs/components/checkbox-chip.svx
```diff
@@ -0,0 +1,107 @@
...
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the component. Falsy values disregarded. |
+| **`inputClass`** | `null` | `string` | A class string to add to the `<input>` element. |
```
aabounegm, line 77: Why does this one not say "Falsey values disregarded." like the rest? If it's because it doesn't disregard them, how about making it consistent with them? And then perhaps move this comment outside the table since it is common to all entries

docs/src/routes/docs/components/checkbox-chip-group.svx
```diff
@@ -0,0 +1,101 @@
...
+    { value: 1.5, label: 'option 1.5', disabled: true },
+    { value: 2, label: 'option 2' },
+  ];
+</script>
+
+# CheckboxChipGroup
+
+A set of options presented as chips with multiple options to select.
+
+If you need to allow selecting only one option, consider using [==`RadioChipGroup`==](./docs/components/radio-chip-group). If you need more customization for the chips, such as their slots, take a look at [==`CheckboxChip`==](./docs/components/checkbox-chip), the building block of this component.
```
aabounegm, line 31: > If you need more customization for the chips, such as their slots...

Maybe we can add a `<slot name="item">` to this component that passes it along to each chip

docs/src/routes/docs/components/checkbox-chip-group.svx
```diff
@@ -0,0 +1,101 @@
...
+## Properties {#properties}
+
+Any properties of `CheckboxChip` can be specified – they will be passed down.
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| ==**`items`**==&nbsp;<span class="required">*</span> |  | `Array` | An array of `{ value, label, disabled, checked }` objects, where only the value is required. Must contain at least one value. |
+| ==**`name`**==&nbsp;<span class="required">*</span> |  | `string` | The name of this checkbox group. |
```
aabounegm, line 70: How about a note specifying that it should be unique from other groups? Or even a link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefname

docs/src/routes/docs/components/checkbox.svx
illright: Yeah, it will be stringified for the DOM, but Svelte doesn't care, and this means that objects can be used as values and retrieved from event values intact

docs/src/routes/docs/components/checkbox-chip.svx
illright: Yeah, that's currently a problem. Not all of them disregard falsy values, which means we must be explicit about which ones do. I do intend to bring consistency to this though, and then we'll remove it.

docs/src/routes/docs/components/checkbox-chip-group.svx
illright: Why? That's exactly the problem that the individual chips are meant to solve + the ability to lay them out in a manual way

docs/src/routes/docs/components/checkbox-chip-group.svx
illright: It doesn't have to be unique, maybe having two groups with the same name is the developer's intention. As for that link, I really don't see a need to educate users on the DOM basics

docs/src/routes/docs/components/checkbox.svx
aabounegm: Yes, but it will cause a problem with the internal value in the HTML (for example, when using it in a form). The JS code can work fine, but the server will receive a bunch of `[Object object]`s

docs/src/routes/docs/components/checkbox-chip-group.svx
aabounegm: This is something in between. You want the features given by the Group component (the implicit `{#each}`, the styles, ...), but also want to customize the content of all chips at once.
I'm not saying it is for this PR, but it is worth considering for later

docs/src/routes/docs/components/checkbox-chip-group.svx
aabounegm: I know it doesn't have to be, but the developer may not be aware of what this prop really means and the implications of reusing the same `name`

docs/src/routes/docs/components/checkbox.svx
illright: True, but, imo, too bad. The developers should know what they're signing up for when they're using HTML forms, and if they are, ensure that only strings get passed here. Don't see a reason to limit everyone else who doesn't use the HTML forms

docs/src/routes/docs/components/checkbox-chip-group.svx
illright: There are barely any styles for the group, so individual components aren't losing much. The `{#each}`, pretty sure every Svelte developer knows how to write their own loop.
I don't like slots for group components like this one

docs/src/routes/docs/components/checkbox.svx
aabounegm: It's not even a limitation, but just a documentation of the expected property type. We expect it to be a string since we pass it to an element that expects it to be a string. If they wanna act smart and use an object, they should know what they are doing. Otherwise, this will confuse others who know that a `value` has to be a string, and they would try to understand how it can be something else and go through the HTML documentation and a big hassle. Let's just not confuse 99% of the users?

PR #97, opened 2020-07-18
Add documentation for the Dialog component

aabounegm:


docs/src/routes/docs/components/dialog.svx
```diff
@@ -0,0 +1,101 @@
+---
+name: Dialog
+shadowRaised: '0 2px 4px -1px transparentize(black, .8), 0 3px 5px 0 transparentize(black, .86), 0 1px 10px 0 transparentize(black, .88);'
```
illright, line 3: This should have actual values, as it won't undergo any preprocessing. Copy it from https://github.com/illright/attractions/blob/develop/docs/src/routes/docs/components/button.svx#L4

docs/src/routes/docs/components/dialog.svx
```diff
@@ -0,0 +1,101 @@
...
+<Showcase>
+  <div slot="showcase" class="padded">
+    <Dialog title="Are you sure you want to exit?" danger closeCallback={() => console.log('cancelled')}>
+      <div slot="title-icon">
+        <AlertTriangleIcon size="25" />
+      </div>
+      <p>You have not saved your changes yet.</p>
+      <p>If you exit without saving, changes will be lost.</p>
+      <p>Are you sure you want to exit?</p>
+      <Button>Yes, exit</Button>
```
illright, line 27: The implicit style guidelines suggest that we write the button texts all lowercase, so let's stick to that in the docs

docs/src/routes/docs/components/dialog.svx
```diff
@@ -0,0 +1,101 @@
...
+    </Dialog>
+  </div>
+  <div slot="source">
+
+```svelte
+<Dialog
+  title="Are you sure you want to exit?"
+  closeCallback={() => alert('cancelled')}
```
illright, line 35: `alert` here, but `console.log` in the actual code. I'd rather have it both be `console.log`, it's less interruptive

docs/src/routes/docs/components/dialog.svx
```diff
@@ -0,0 +1,101 @@
...
+</Showcase>
+
+
+## Properties {#properties}
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`danger`** | `false` | `boolean` | Applies the danger color to the title (including the icon slot). |
```
illright, line 58: Maybe we should mention this only applies if the icon slot contains an `<svg>`?

aabounegm: How come? I don't see this limitation anywhere in the SCSS code

illright: Well, for one, only `<svg>` can be customized at all, for two-

I was about to show you the line of code only to realize that it's not there and we rely solely on the fact that feather icons have `stroke="currentColor"` by default. Maybe we should add a rule for that `svg` and force `stroke: currentColor`

aabounegm: > only `<svg>` can be customized at all

not really. this is a slot, which means any content can be supplied by the user, including anything that is styled to respond to the `.danger` class

> Maybe we should add a rule for that `svg` and force `stroke: currentColor`

Are you talking about the docs or the implementation? I'm getting a bit confused

illright: I mean that `` doesn't respond to CSS. As for that rule addition, I mean to add it to the implementation, so that we live up to the promise that an `<svg>` icon responds to the `danger` prop

PR #99, opened 2020-07-19
Add documentation for the Dot component

aabounegm:


docs/src/routes/docs/components/dot.svx
```diff
@@ -0,0 +1,86 @@
...
+</script>
+
+# Dot
+
+A dot.
+
+<Showcase>
+  <div slot="showcase" class="padded">
+    <Button>
+      Plain ol' dot: <Dot />
```
illright, line 18: I know that with this thing you meant to demonstrate that dots don't have margin by default, but it's such an eye-sore, so I'd rather you added margin there as well

docs/src/routes/docs/components/dot.svx
```diff
@@ -0,0 +1,86 @@
...
+## Properties {#properties}
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`info`** | `false` | `boolean` | Applies the `$info` color to the dot |
+| **`attention`** | `false` | `boolean` | Applies the `$attention` color to the dot |
+| **`danger`** | `false` | `boolean` | Applies the `$danger` color to the dot |
+| **`success`** | `false` | `boolean` | Applies the `$success` color to the dot |
```
illright, line 66: We should probably mention that all these color props are mutually exclusive, kind of like it's done in the button. Listing them all in each cell would be too much and too repetitive, so how about we write this in a paragraph right above the table? Don't forget the `==`, to remain consistent with the Button, when listing those props in that paragraph

docs/src/routes/docs/components/dot.svx
```diff
@@ -0,0 +1,86 @@
...
+
+<script>
+  import { Dot, Button } from 'attractions';
+  import Showcase from 'src/containers/docs/showcase.svelte';
+  import ColorPreview from 'src/components/docs/color-preview.svelte';
+</script>
+
+# Dot
+
+A dot.
```
illright, line 13: You, sir, get an F for expressiveness)
Maybe we could play around with this, something like "a colorful dot to signal about all your important events"

docs/src/routes/docs/components/dot.svx
```diff
@@ -0,0 +1,86 @@
...
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`info`** | `false` | `boolean` | Applies the `$info` color to the dot |
+| **`attention`** | `false` | `boolean` | Applies the `$attention` color to the dot |
+| **`danger`** | `false` | `boolean` | Applies the `$danger` color to the dot |
+| **`success`** | `false` | `boolean` | Applies the `$success` color to the dot |
+| **`small`** | `false` | `boolean` | Decreases the size of the dot |
+
+### Class Properties
```
illright, line 69: Don't forget `restProps`. Yes, that will mean creating a _Functional properties_ heading just for this one entry, but consistency above all

PR #101, opened 2020-07-20
Document the file input components

illright:


docs/src/routes/docs/components/file-dropzone.svx
```diff
@@ -0,0 +1,95 @@
...
+## Events {#events}
+
+| Name | Event Detail | Description |
+| ---- | ------------ | ----------- |
+| **`change`** | `{ value, nativeEvent }` | Fired when a selection is added to/removed from. In case of removal, the `nativeEvent` isn't present in the event detail. |
+
+
+## Slots {#slots}
+
+### `empty-layer` slot
```
aabounegm, line 75: Maybe we should expose the `disabled`, `wrongType`, and `dragActive` variables to this slot

illright: Good call

docs/src/routes/docs/components/file-dropzone.svx
```diff
@@ -76,6 +76,11 @@ All `<FileDropzone>` components use the `.file-dropzone` class.
 
 The inner content of the dropzone when no files are selected. Defaults to a message that responds to the state of the dropzone: empty, with files ready to be dropped, with incorrect files ready to be dropped or disabled.
 
+| Prop Name | Description |
```
aabounegm, line 79: How about adding `Type` column instead of having to state "The boolean state ..." in the description?

illright: Reasonable

PR #102, opened 2020-07-21
Document the FormField component

illright:


docs/src/routes/docs/components/form-field.svx
```diff
@@ -0,0 +1,121 @@
...
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`name`** | `null` | `string` | The name of the form field. Displayed prominently next to the actual field. |
+| **`help`** | `null` | `string` | The subtitle text under the name providing extra guidance. |
+| **`id`** | `null` | `string` | The ID to add the the `for` attribute of the `<label>` element containing the ==`name`==. Useful if you pass an ID to the actual field inside and want to connect it with the label. |
+| **`required`** | `false` | `boolean` | Marks the form field as required, adding an asterisk to the name and text under the field indicating that the field is mandatory. |
+| **`optional`** | `false` | `boolean` | Marks the form field as optional, explicitly stating the optionality of the field in the text under the field. |
+| **`errors`** | `null` | `Array` | An array of error messages to display under the field. Falsy values disregarded. |
```
aabounegm, line 85: For similar reasoning as the `null` vs `Infinity` before, isn't `[]` a more reasonable default value here?

docs/src/routes/docs/components/form-field.svx
```diff
@@ -0,0 +1,121 @@
...
+
+### Class Properties
+
+All `<FormField>` components use the `.form-field` class.
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the component. Falsy values disregarded. |
+| **`nameClass`** | `null` | `string` | A class string to the `<label>` containing the form field name. Falsy values disregarded. |
+| **`helpClass`** | `null` | `string` | A class string to the help text of the form field. Falsy values disregarded. |
```
aabounegm, line 95: Missing a verb after `a class string to...`

docs/src/routes/docs/components/form-field.svx
```diff
@@ -0,0 +1,121 @@
...
+The content of the form field – the actual field itself.
+
+### `description` slot
+
+The block describing the form field, which by default is taken up by the ==`name`== and ==`help`==. If you need a custom description of the field, use this slot.
+
+### `message` slot
+
+Additional text to put underneath the actual field along with the messages like "required", "optional" and errors.
+
```
aabounegm, line 111: I don't really get why does the first slot replace the content, while the second one just appends to it

docs/src/routes/docs/components/form-field.svx
illright: I must've copied it from somewhere where there still is this problem

docs/src/routes/docs/components/form-field.svx
illright: Because it's easier that way. The first slot is something very rarely desired, so we can afford to tear down the content there

docs/src/routes/docs/components/form-field.svx
illright: Yeah, probably

PR #103, opened 2020-07-21
Make the tiles on the home page capable of navigation

illright:
The other day I was told (and witnessed how someone kept pressing the tiles) that the tiles look like something you should click, so they completely overlooked the buttons (through the mobile version, anyway).

So I figured, why not. The links for the first two tiles are kind of best-fit, I think it's worth having a separate page listing all the components in the future, which can be an optimal destination for the first tile, and we should also have a separate list of all SCSS variables, where the second tile will be linking. In the meantime,

### DIFF
--- a/docs/src/containers/home/info-tiles.svelte
+++ b/docs/src/containers/home/info-tiles.svelte
@@ -12,16 +12,19 @@
     icon={GridIcon}
     title="{totalComponents} component{s(totalComponents)}"
     subtitle="and more to come!"
+    href="./docs/components/button"
   />
   <InfoTile
     icon={Edit2Icon}
     title="Stylable with Sass"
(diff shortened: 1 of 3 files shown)


docs/src/containers/home/info-tiles.svelte
```diff
@@ -12,16 +12,19 @@
     icon={GridIcon}
     title="{totalComponents} component{s(totalComponents)}"
     subtitle="and more to come!"
+    href="./docs/components/button"
```
aabounegm, line 15: Maybe it is worth creating an `index.svx`, even if with something as simple as "check the components on the left"

illright: Nah, for now, the button serves better. Later we will create a proper page

PR #104, opened 2020-07-21
Add documentation for Dropdown components

aabounegm:


illright:
I beg your pardon
image

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,98 @@
...
+| **`open`** | `false` | `boolean` | Controls whether the dropdown content is shown or hidden. |
+
+### Class Properties
+
+All `<DropdownShell>` components use the `.dropdown-shell` class
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the underlying `<div>` element. |
+
```
illright, line 86: Mention here that one can take advantage of the `dropdown-chevron` class, like it's done in https://illright.github.io/attractions/104/docs/components/accordion-section

aabounegm:
To demonstrate this:
image

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,98 @@
...
+    <DropdownShell>
+      Useless without a <code>&lt;Dropdown /&gt;</code>
+    </DropdownShell>
+  </div>
+  <div slot="source">
+
+```svelte
+<DropdownShell let:toggle>
+  <Button on:click={toggle}>
+    Can contain more than one
```
illright, line 49: Oh god, no, let's not endorse that. This is uninteded behaviour and I see 0 logical reasons for it to be used

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,106 @@
...
+<RelatedComponents components={relatedComponents} />
+
+<Showcase>
+  <div slot="showcase" class="padded">
+    <DropdownShell let:toggle>
+      <Button on:click={toggle}>
+        I toggle a dropdown
+      </Button>
+      <Dropdown>
+        <div class="padded">
```
illright, line 31: This is not enough, they still look crappy. Please, add a SCSS file and give them some padding on all sides.

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,106 @@
...
+</Showcase>
+
+
+## Properties {#properties}
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`right`** | `false` | `boolean` | Aligns the dropdown content such that it opens to the right of the shell |
```
illright, line 88: Not to the right, but aligned with the right edge

illright: Also don't forget periods

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,106 @@
...
+| **`top`** | `false` | `boolean` | Aligns the dropdown content such that it opens to the top of the shell |
+
+### Class Properties
+
+All `<Dropdown>` components use the `.dropdown` class
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the underlying `<div>` element. |
+
```
illright, line 98: The default slot should be documented

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,106 @@
...
+    </DropdownShell>
+    <Dropdown>
+      Don't use it without its shell
+    </Dropdown>
+  </div>
+  <div slot="source">
+
+```svelte
+<DropdownShell let:toggle>
+  <Button on:click={toggle}>
```
illright, line 54: Also show the chevron with the `dropdown-chevron` class

illright: aaaand the buttons are all lowercase

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,106 @@
...
+</Dropdown>
+```
+
+  </div>
+</Showcase>
+
+
+## Properties {#properties}
+
+### Functional Properties
```
illright, line 84: They actually kinda look like style properties to me

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,98 @@
...
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the underlying `<div>` element. |
+
+
+## Events {#events}
+
+| Name | Event Detail | Description |
+| ---- | ------------ | ----------- |
+| **`change`** | `value: boolean` | Fired when toggling the dropdown (with `value` being `true` for "open" and `false` for "closed"). |
```
illright, line 92: All the previous doc pages didn't have type hints, and they'd be too space-consuming anyway. I think the description sums up the value well.

Also now that I think about it, we should probably make this consistent with the other events and wrap it in an object with a `value` property.

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,98 @@
...
+
+## Functions {#functions}
```
illright, line 94: New section, huh. I think it's best to name it `Exported Attributes` or `Exposed Attributes` and have the columns `Name | Type | Description`. Not all such attributes will be functions, and we need to accomodate for them. Arguments can be described in the description.

And in here I'd also like to mention that one would wanna use this attribute when toggling must be done in the same component where the shell is created. To put that in the description.

Also it might be worth adding under the header that those attributes can be accessed by using `bind:this` to a variable.

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,98 @@
...
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the underlying `<div>` element. |
+
+
+## Events {#events}
+
+| Name | Event Detail | Description |
+| ---- | ------------ | ----------- |
+| **`change`** | `value: boolean` | Fired when toggling the dropdown (with `value` being `true` for "open" and `false` for "closed"). |
+
```
illright, line 93: To document slots too, in here it's especially important since that slot has a prop

illright:
Yeah, I get that you wanted to stress that it's bad to use the dropdown without a shell, but let's just leave this for the users to try on their own and not break our own pages

docs/src/routes/docs/components/dropdown-shell.svx
aabounegm: Just wanted to show different content than in the Dropdown and got a bit creative)

docs/src/routes/docs/components/dropdown.svx
aabounegm: I just copied it from the demo. It was enough there.

docs/src/routes/docs/components/dropdown-shell.svx
aabounegm: Yeah, the only reason I wrote its type is that it wasn't in an object so I thought it doesn't look very clear

docs/src/routes/docs/components/dropdown-shell.svx
aabounegm: I think a better way of stating the arguments (and return) types would be something similar to the Typescript way: `() => void` (omitting arg names as they are not relevant), but I'll stick with `Function` for now for consistency with the rest of the docs.

illright: Actually, we have some sort of internal inconsistency, because there's also the `async function*` as the type in the Autocomplete. Maybe we could have written something like AsyncGenerator, which, I believe, really is the type, but imo that one's very confusing. In this way, we should probably find a better way to specify functions. If the TS way would allow us to express async generators, then yes, that'd be the superior way

attractions/dropdown/dropdown-shell.svelte
```diff
@@ -8,7 +8,7 @@
 
   export function toggle() {
     open = !open;
-    dispatch('change', open);
+    dispatch('change', { value: open});
```
illright, line 11: Don't forget a space on both sides of the braces

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -0,0 +1,112 @@
...
+
+
+## Properties {#properties}
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`right`** | `false` | `boolean` | Aligns the dropdown content with the right edge of the shell. |
+| **`top`** | `false` | `boolean` | Aligns the dropdown content with the top edge of the shell. |
```
illright, line 86: I actually kinda think the previous version sounded better over here, saying that the dropdown is to the top of the shell. This can be interpreted as if the dropdown starts at the top edge of the shell, but still extends down

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,104 @@
...
+| Name | Event Detail | Description |
+| ---- | ------------ | ----------- |
+| **`change`** | `{ value }` | Fired when toggling the dropdown (with `value` being `true` for "open" and `false` for "closed"). |
+
+
+## Slots {#slots}
+
+### Default slot
+
+The shell of the dropdown. The dropdown itself must also be inside it since it exposes the `toggle` function as a slot prop.
```
illright, line 93: but..but.. we have a table for slot props...

attractions/dropdown/dropdown-shell.svelte
aabounegm: I felt something was wrong but couldn't quite place it 😅

docs/src/routes/docs/components/dropdown.svx
aabounegm: Do you mean that the previous comment applied only to the `right` prop (and it is better now), but not to the `top`?

illright: Yeah, the `top` was fine as you initally wrote it

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,108 @@
...
+
+### Class Properties
+
+All `<DropdownShell>` components use the `.dropdown-shell` class
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`class`** | `null` | `string` | A class string to add to the underlying `<div>` element. |
+
+==Bonus==: use the `.dropdown-chevron` class for the chevrons on the dropdown shell to get a nice rotation animation (see showcase).
```
illright, line 80: More like "on the dropdown handle", but I'm not sure if the term "handle" will be obvious to everyone. But having the "chevrons on the dropdown shell" doesn't sound right.

Maybe we could use "triggering button" or "triggering handle" or some other phrasing

docs/src/routes/docs/components/dropdown-shell.svx
```diff
@@ -0,0 +1,108 @@
...
+    <DropdownShell>
+      Useless without a <code>&lt;Dropdown /&gt;</code>
+    </DropdownShell>
+  </div>
+  <div slot="source">
+
+```svelte
+<DropdownShell let:toggle>
+  <Button on:click={toggle}>
+    this is the dropdown shell
```
illright, line 46: Which brings me over here, actually. If we say here "this is the dropdown handle" instead of the shell (because it isn't the shell, the shell is around them both), we can get away with using the "handle" term in the previous comment

docs/src/routes/docs/components/dropdown.svx
illright: The demo was inherently crappy :)

PR #109, opened 2020-07-23
Document the Pagination component

illright:


docs/src/routes/docs/components/pagination.svx
```diff
@@ -0,0 +1,71 @@
+---
+name: Pagination
+shadow0: '0 2px 2px 0 rgba(0, 0, 0, .14), 0 1px 5px 0 rgba(0, 0, 0, .12);'
+shadow1: '0 2px 5px rgba(0, 0, 0, .14), 0 1px 5px rgba(0, 0, 0, .12);'
+shadowRaised: '0 2px 4px -1px rgba(0, 0, 0, .2), 0 3px 5px 0 rgba(0, 0, 0, .14), 0 1px 10px 0 rgba(0, 0, 0, .12);'
```
aabounegm, line 5: Where are these being used?

illright:
As for the arrows, I don't think this could be fixed easily. We're listening to `change` events

aabounegm:
@illright Couldn't we listen to input events instead and check the pressed key manually?

illright:
I tried listening to the `input` events, it turns out that either HTML or Svelte doesn't fire `input` events for Enter key presses on `type="number"` inputs, so I really don't know how else to implement this. And honestly, it doesn't matter too much to me. We don't show spinners for those inputs anyway, and using the keyboard to navigate random access is strange, imo. So I'm not too worried

aabounegm:
Neither `input` nor `change` get fired for Enter key. The simplest solution would be to listen to `keydown` instead and only react to Enter key presses.

aabounegm:
Also, I'm not sure why you said we are not showing spinners. They are showing for me:
image

PR #112, opened 2020-07-24
Document the Popover component

illright:


docs/src/routes/docs/components/popover.svx
```diff
@@ -0,0 +1,95 @@
...
+### `PopoverPositions` {#popover-positions}
+
+An enum describing possible positions for the `Popover`.
+
+| Field | Description |
+| ---- | ----------- |
+| **`TOP`** | Above the handle, centered with respect to it. |
+| **`BOTTOM`** | Below the handle, centered with respect to it. |
+| **`LEFT`** | To the left of the handle, centered with respect to it. |
+| **`RIGHT`** | To the right of the handle, centered with respect to it. |
```
aabounegm, line 95: I don't think you emphasized enough that it is centered with respect to it)

illright: I wish there was a way around this)

PR #113, opened 2020-07-24
Document the Snackbar components

illright:


docs/src/routes/docs/components/snackbar-container.svx
```diff
@@ -0,0 +1,114 @@
...
+An enum listing the possible positions for the snackbar stack.
+
+| Field | Description |
+| ---- | ----------- |
+| **`TOP_LEFT`** | In the top left corner. |
+| **`TOP_MIDDLE`** | At the top, horizontally centered. |
+| **`TOP_RIGHT`** | In the top right corner. |
+| **`MIDDLE_RIGHT`** | Along the right edge, vertically centered. |
+| **`BOTTOM_RIGHT`** | In the bottom right corner. |
+| **`TOP_MIDDLE`** | At the bottom, horizontally centered. |
```
aabounegm, line 100: Careful with the copy-paste

docs/src/routes/docs/components/snackbar-container.svx
```diff
@@ -0,0 +1,114 @@
...
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`position`** | `SnackbarPositions.BOTTOM_LEFT` | [==`SnackbarPositions`==](#snackbar-positions) | The position of the snackbar stack inside the window. |
+
+### Class Properties
+
+All `<SnackbarContainer>` components create the `.snackbar-stack` element after the default slot content.
```
aabounegm, line 62: `.snackbar-stack` is not an element name

docs/src/routes/docs/components/snackbar.svx
```diff
@@ -0,0 +1,89 @@
...
+
+## Properties {#properties}
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| ==**`text`**==&nbsp;<span class="required">*</span> |  | `string` | The text to show on the snackbar. |
+| ==**`closeCallback`**==&nbsp;<span class="required">*</span> |  | `Function` | The callback to call to close the snackbar in the `SnackbarContainer`. Passed internally by the `SnackbarContainer` when using the `showSnackbar` function. |
+| **`action`** | `null` | `{ text, callback() }` | The action for the button on the snackbar. If this is `null`, the button is not rendered. Otherwise it has `text` as a label and calls `callback` on click. |
```
aabounegm, line 61: We usually don't mark the types of object properties, why did you feel the need to write the `()` in `callback()` here?

docs/src/routes/docs/components/snackbar.svx
```diff
@@ -0,0 +1,89 @@
...
+| **`textClass`** | `null` | `string` | A class string to add to the label of the snackbar. |
+| **`buttonClass`** | `null` | `string` | A class string to add to the action button of the snackbar. |
+
+
+## SCSS Variables {#scss-variables}
+
+| Name | Description | Default |
+| ---- | ----------- | ------- |
+| **`$snackbar-bg`** | The background of the snackbar. | <ColorPreview value="#404040" /> |
+| **`$snackbar-fg`** | The text color of the snackbar. | <ColorPreview value="#404040" /> |
```
aabounegm, line 84:
```suggestion
| **`$snackbar-fg`** | The text color of the snackbar. | <ColorPreview value="#FFFFFF" /> |
```

docs/src/routes/docs/components/snackbar-container.svx
illright: It's a CSS selector, I just figured that'd be enough. Any suggestions?

docs/src/routes/docs/components/snackbar.svx
illright: This is a very ugly table, honestly. Originally I wanted to have this cell say `{ text: string, callback: Function }`, but this is a huge blob of unwrappable text, and unfortunately, the default value column also had a huge blob of unwrappable text, and the whole table just went to shit, so I figured I'd at least indicate that the callback is a function without explicitly writing out the types

illright:
You're welcome to give a shot at simplifying this component)

docs/src/routes/docs/components/snackbar-container.svx
aabounegm: It is understandable, just not really technically accurate. "... create an element with the `.snackbar-stack` class after ..." would be more accurate. (Frankly, I don't like that we write both the dot and the word "class", but we do that everywhere and it is understandable enough)

docs/src/routes/docs/components/snackbar.svx
aabounegm: I am all for specifying the types everywhere, but perhaps we need a better format for the table. For now, we can wrap the object (although that might be too many lines for each row) to look something like:
```typescript
{
    text: string,
    callback: () => void
}
```
It would be great if we use the Typescript notation, and even greater if we add type definitions file

illright: We'll hold on to full TypeScript notation until after we're done, but for now I did the wrapping

PR #116, opened 2020-07-27
Tabs documentation

aabounegm:


attractions/tab/tab.scss
```diff
@@ -32,6 +32,10 @@
     }
   }
 
+  > input:disabled + .content {
+    cursor: not-allowed;
```
illright, line 36: I think it would also help to give it `color: $disabled`

docs/src/routes/docs/components/tab.svx
```diff
@@ -0,0 +1,98 @@
...
+</Showcase>
+
+
+## Properties {#properties}
+
+### Style Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`ripple`** | `true` | `boolean` | Adds a ripple to the tab when clicking on it. |
```
illright, line 53: Perhaps we should invert this prop for consistency with the rest of the library, to `noRipple` instead

docs/src/routes/docs/components/tab.svx
```diff
@@ -0,0 +1,98 @@
...
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| ==**`value`**==&nbsp;<span class="required">*</span> |  | `string` | The value assigned to the `<input type="radio">`. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Value) for more information. |
+| **`group`** | `null` | `string` | The name used to group several `<Tab>`s together. |
```
illright, line 60: Wait, no, that's not what `group` is, that's `name`. `group` is something you can bind to to get the value of the currently selected radio button in the name group

illright: Which makes me wonder why are we not explicit about having the `name` prop with the tabs. I think we should be

aabounegm: We are explicit about having the `name` prop with the tabs. Check https://github.com/illright/attractions/blob/docs/tabs/docs/src/routes/docs/components/tabs.svx#L51

illright: Yeah, but not the `Tab`, which is strange. Best to be explicit everywhere

PR #117, opened 2020-07-27
Request bundle size from BundlePhobia

aabounegm:
The horrible formatting is needed to remove extra space before the comma

### DIFF
--- a/docs/src/containers/home/info-tiles.svelte
+++ b/docs/src/containers/home/info-tiles.svelte
@@ -1,10 +1,17 @@
 <script>
+  import { onMount } from 'svelte';
   import * as attractions from 'attractions';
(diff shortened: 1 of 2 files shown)


illright:
Don't forget about the homepage too, we also show the file size there

illright:
Also maybe we should do this at build time? Makes more sense to me, since this is a static value

aabounegm:
Oh, so the "Please help me" text was actually for me to fill? I thought it was for the user to click.

I also thought about having it at build time, but then I realized that the docs are built simultaneously as package is being published to npm (and will probably be done before it). And even if we can guarantee publishing to npm before building the docs, who knows how long it takes for bundlephobia to update its database?

PR #121, opened 2020-08-02
Document typography components

illright:


docs/src/routes/docs/components/typography.svx
```diff
@@ -0,0 +1,89 @@
...
+  import { H1, H2, H3, Headline, Subhead, Label } from 'attractions';
+  import Showcase from 'src/containers/docs/showcase.svelte';
+  import ColorPreview from 'src/components/docs/color-preview.svelte';
+</script>
+
+# Typography
+
+Typography elements with embedded responsive styles.
+
+Also comes with a `<TypographyStyles>` component, which only adds global styles, setting the `$font` for the whole page, coloring links, etc. A good place to put it for Sapper, for example, would be anywhere in the main [layout file](https://sapper.svelte.dev/docs#Layouts).
```
aabounegm, line 15: I think an example of this `<TypographyStyles>` component has to be shown. I cannot understand its purpose or how to use it just from this page

docs/src/routes/docs/components/typography.svx
```diff
@@ -0,0 +1,89 @@
...
+
+### Functional Properties
+
+| Name | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
+| **`...`** |  | `any` | Everything will be passed to the underlying element. |
+
+### Class Properties
+
+_The following only applies to `<Headline>`, `<Subhead>` and `<Label>` components._
```
aabounegm, line 65: Not a fan of the Oxford comma, I see)

docs/src/routes/docs/components/typography.svx
```diff
@@ -0,0 +1,89 @@
...
+
+All `<Headline>` components use the `.headline` class, all `<Subhead>` components use the `.subhead` class, all `<Label>` components use the `.label` class.
```
aabounegm, line 67: The grammar police is issuing a warning for the run-on sentence)
```suggestion
All `<Headline>` components use the `.headline` class, all `<Subhead>` components use the `.subhead` class, and all `<Label>` components use the `.label` class.
```

docs/src/routes/docs/components/typography.svx
```diff
@@ -0,0 +1,89 @@
...
+
+## SCSS Variables {#scss-variables}
+
+| Name | Description | Default |
+| ---- | ----------- | ------- |
+| **`$main`** | The color of some headings, the headline and the label. <br /> When `<TypographyStyles>` are applied, also colors the links. | <ColorPreview value="#4300B0" /> |
+| **`$link-visited`** | When `<TypographyStyles>` are applied, colors the visited links. | <ColorPreview value="#50005F" /> |
+| **`$bold-font-weight`** | The bold font weight of headings. <br /> When `<TypographyStyles>` are applied, also adjusts the boldness of `<strong>`. | `500` |
+| **`$thin-font-weight`** | The thin font weight of the subhead. | `300` |
+| **`$font`** | When `<TypographyStyles>` are applied, becomes the font for the entire page. | `'Ubuntu'` |
```
aabounegm, line 89: As mentioned in a previous comment, it is not clear _"When `<TypographyStyles>` are applied"_ really means

docs/src/routes/docs/components/typography.svx
illright: Russian upbringing surfaces

docs/src/routes/docs/components/typography.svx
illright: Sigh. Yeah, this is a major weakness. For some reason I actually feel a lot of dissatisfaction with those docs in general, but I do not yet know how to improve. In this case as well, I don't know how to improve this. Honestly, the whole idea of the `TypographyStyles` component kinda blows, and I believe we didn't even use it in Innopoints, so maybe it could just be removed altogether

docs/src/routes/docs/components/typography.svx
illright: Idk how we can improve this

docs/src/routes/docs/components/typography.svx
aabounegm: by _"those docs in general"_, are you referring to the typography docs or all components' docs? If all of them, why so?
About the `<TypographyStyles>` component (of whose existence I was previously unaware), I see that all it does is include some styles. They are not even that many styles and I think we can let go of it entirely. However, if you believe it might be useful, we can export the SCSS file itself directly such that users can include it in their project as they want (in their own SCSS or in the preprocessor's `includePaths`) as it is not really a component, but simply a bunch of styles made `:global` (imo, the purpose of a component should not be for side-effects purely, if at all)

docs/src/routes/docs/components/typography.svx
aabounegm: Technically, both are correct, but I side with the Oxford comma team)

docs/src/routes/docs/components/typography.svx
illright: All of them. [Here's a framework](https://ant.design/components/button/), for example, that has lots of examples in a rather easy to consume format. Originally I had a lot of faith in the showcase section, but it has proven to be unsuitable in the general case, and not even useful. Examples in that framework, for instance, show actual real-world usage of the components, and the showcase is a very artificial, dead way of displaying them, aiming to showcase as much stuff as possible in as little space, which harms it even more. The prop tables and generally the structure of the documentation pages also feels dead to me, like we're taking consistency over clarity, which makes the docs relentlessly consistent but hard to read and learn from, makes them more machine than human.

As for TypographyStyles, yes, fine, let's just remove this component.

aabounegm: Don't beat yourself too hard over this. I think they look great and easy to understand, and our components are simple enough that we do not need that much examples. I think we should just go with it and wait for any feedback

PR #124, opened 2020-08-10
Feature/file dropzone max

aabounegm:
Adds a prop that limits the number of files allowed in the dropzone


illright:
Since you updated Sapper, we should probably go over https://sapper.svelte.dev/migrating/

illright:
Also please add `display: flex` to `.copyable > [slot="source"]` in the styles for the Showcase. Perhaps this broke from when we bumped prism, or it was always broken, but this will ensure that the horizontal scrollbar in the code section of the showcase is all the way at the bottom and not awkwardly halfway through

illright:
And I think we do have to change something for the service worker as per the migration guide because now when I open the FileDropzone page and reload, I get incorrect styling and `Request failed` from the service worker in the console

attractions/file-input/file-dropzone.svelte
```diff
@@ -106,11 +107,13 @@
     on:dragenter={() => dragActive = true}
     on:dragleave={() => { dragActive = false; wrongType = false; }}
     on:drop|preventDefault|stopPropagation={acceptUpload}
-    use:ripple={{ disabled }}
+    use:ripple={{ disabled: disabled || files.length >= max }}
```
illright, line 110: For some reason, the ripple is still present when I hit the limit of files

docs/src/routes/docs/components/file-dropzone.svx
```diff
@@ -51,6 +51,7 @@ For a simpler file input without the built-in representation of the file selecti
 | **`files`** | `[]` | `Array` | The user's selection as an array of [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) objects. Note: this is an actual `Array`, not a `FileList`, as opposed to the native `<input type="file">`. |
 | **`beforeChange`** | `null` | `Function` | A callback to call for each file that passes the ==`accept`== check before it is added to the ==`files`==. <br /> If it returns a Promise, they will be started for every file, but awaited at the end. |
 | **`disabled`** | `false` | `boolean` | Whether the input should accept files. |
+| **`max`** | `Infinity` | `number` | The maximum number of allowed files. |
```
illright, line 54: I don't really like the `allowed` here. Maybe `The maximum amount of files that one can upload`?

attractions/file-input/file-dropzone.svelte
aabounegm: This is because we do not react to changes in its parameters. The same would apply to, say, buttons if they are disabled after being previously enabled (though we wouldn't see that because the button itself is disabled, so no `click` event is fired).

aabounegm:
You are right about the migration guide, I didn't notice it when upgrading. The service worker is one side effect of this, as they updated the `files` array to be prefixed with '/', which broke our relative base URL thing.
Also, I don't really get what "broke" about the style, but fixed anyways.

docs/rollup.config.js
```diff
@@ -20,10 +20,10 @@ const mode = process.env.NODE_ENV;
 const dev = mode === 'development';
 const legacy = !!process.env.SAPPER_LEGACY_BUILD;
 
-const onwarn = (warning, onwarn) => (
-  warning.code === 'CIRCULAR_DEPENDENCY'
-  && /[/\\]@sapper[/\\]/.test(warning.message)
-) || onwarn(warning);
+const onwarn = (warning, onwarn) =>
+  (warning.code === 'MISSING_EXPORT' && /'preload'/.test(warning.message)) ||
```
illright, line 24: I kinda prefer keeping the operator in the beginning of the next line as opposed to the end of the current one. Easier to read, because by the time your eyes latch on to a new line, you'll forget what the operator was

aabounegm: I took it directly from the template, and it helps in alignment, but fine.

PR #127, opened 2020-08-20
Add the utils documentation page

aabounegm:


docs/src/routes/docs/_layout.svelte
```diff
@@ -132,6 +132,10 @@
         },
       ],
     },
+    {
+      title: 'Utilities',
+      segment: 'utils',
```
illright, line 137: A mismatch between "utilities" in the left nav and "utils" in the breadcrumbs

docs/src/routes/docs/utils.svx
```diff
@@ -0,0 +1,21 @@
...
+---
+
+# Utility functions
+
+A collection of helper functions that can be imported by name from `attractions/utils`. For example: `import { classes } from 'attractions/utils';`
+
+
+| Name | Signature | Description |
+| ---- | --------- | ----------- |
+| **`callOnSight`** | `(node, { callback, args }) => { destroy }` | An [action](https://svelte.dev/tutorial/actions) that is intended to be used with the `use` directive. Calls the provided `callback` with the given `args` (spreaded) once the component appears on the screen. |
```
illright, line 12: Yeah, that's what I was fearing the most. This stuff is very hard to read, but I suppose it will do for now

docs/src/routes/docs/_layout.svelte
aabounegm: Because the breadcrumbs component uses the `segment` and not the `title`. Should I change it to use the title instead or do you want the segment (and therefore the URL as well) to use the full word?

illright: Use the full word for the URL

docs/src/routes/docs/utils.svx
aabounegm: I think only functions intended for use with directives (such as `use:`) are the ones difficult to read. I don't really have a better idea how to document them

illright: No, I meant that the description column in the table is squeezed

illright: We should remove the table in the future

PR #136, opened 2020-08-24
Changelog

aabounegm:
Do you think the top 2 lines in the changelog file should be there or not? Should anything be there at all or just jump right to main content?

The format is based on keepachangelog.com, and the "Unreleased" section should only remain as long as the file is in develop. Its purpose is to be edited with every change so that nothing is forgotten and we don't have to struggle to remember all the changes each release. It should get an actual tag in the same commit that bumps the version and prepares for release.

**Issues:**
For some reason, importing the `.md` file inside the `.svx` one causes the YAML metadata defined in the latter to somehow get overwritten with the ones from the former (even if there were none). Unfortunately, GitHub renders this metadata in a table in the top of the page, which is not desirable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
-## [Unreleased] - yyyy-mm-dd
+## [Unreleased]
 
 Just minor documentation fixes.
 
 ### Added
 - [`<Tab>`](https://github.com/illright/attractions/blob/master/attractions/tab/tab.svelte)s
   now also dispatch `click` event.
-
-### Changed
 
 ### Fixed
 - Mobile navigation in documentation ("Components" tab dropdown now works as expected).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ name: Changelog
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
+The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - yyyy-mm-dd
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,32 @@
+---
+name: Changelog
+---
+
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased] - yyyy-mm-dd
+
+Just minor documentation fixes.
+
+### Added
+- [`<Tab>`](https://github.com/illright/attractions/blob/master/attractions/tab/tab.svelte)s
+  now also dispatch `click` event.
+
+### Changed
+
+### Fixed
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
...
+name: Changelog
+---
+
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased] - yyyy-mm-dd
```
illright, line 11: I don't really like the `yyyy-mm-dd`, and it doesn't seem to be in Keepachangelog either, why'd you add it?

CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
...
+
+Just minor documentation fixes.
+
+### Added
+- [`<Tab>`](https://github.com/illright/attractions/blob/master/attractions/tab/tab.svelte)s
+  now also dispatch `click` event.
+
+### Changed
```
illright, line 19: Also how about we omit unused headings?

illright (changes requested):

illright:
I'm fine with the YAML frontmatter in the Changelog.md file

docs/src/routes/docs/changelog.svx
```diff
@@ -1,3 +1,9 @@
 ---
 name: Changelog
 ---
+
+<script>
+  import Changelog from 'root/CHANGELOG.md';
```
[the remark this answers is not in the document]
aabounegm: Yes we can

CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Changelog
+---
+
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
```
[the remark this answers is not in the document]
aabounegm: I copy-pasted it directly from some place on GitHub; didn't notice

CHANGELOG.md
aabounegm: It was in the template I copied from. I thought I'd leave to make clear the format, but I think we both agree that this is the superior format and we have no American developers here)

CHANGELOG.md
aabounegm: I thought of keeping them at least in the unreleased section so that we remember what sections there are, but I guess it should be intuitive enough to remember. At worst we can consult https://keepachangelog.com/ again

aabounegm:
So this looks fine to you?
image
I don't really like it. Its purpose can be confusing for some. Let's not decide on this until we understand why MDsveX does what it does?

illright:
I think we should merge it like this. Best to have a changelog instead of a blank page

aabounegm:
I was waiting on https://github.com/pngwn/MDsveX/issues/130, but it seems to be taking longer than expected. Let's merge this one and then open a new PR after that issue gets fixed.

CHANGELOG.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
+
+## [Unreleased]
+
+Just minor documentation fixes.
```
illright, line 13: Then we should write this up as a release

illright (changes requested):

CHANGELOG.md
aabounegm: I wanted to fix the styles issue as well before releasing a new version

illright (approved):

PR #137, opened 2020-09-01
Docs/hotfix/styles

aabounegm:
When navigating to a docs page directly, the global styles were not loaded because they were only included in the main page.
This should fix it

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -5,3 +5,4 @@ yarn-error.log
 /cypress/screenshots/
 /__sapper__/
 .env
+.svx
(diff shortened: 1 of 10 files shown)


docs/.gitignore
```diff
@@ -5,3 +5,4 @@ yarn-error.log
 /cypress/screenshots/
 /__sapper__/
 .env
+.svx
```
illright, line 8: What is that for?

aabounegm: It was a folder that got generated when I ran `yarn docs:export`

illright: Can't reproduce. It shouldn't generate any folders

aabounegm: Maybe it's a Windows thing, idk

aabounegm: Doesn't hurt, anyway

illright (approved):

PR #140, opened 2020-09-02
Update the dropdown event interface

### LINKED ISSUE #138: Datepickers in demo page do not always open
First thanks for you library. I have started using it on a small project. While testing out the datepicker (in Chrome version 84.0.4147.135 on a mac), I noticed the datepicker does not always open if you click between the four pickers a number of times.

illright:
Previously the DatePicker was built with a `bool` event detail, which was changed to `{ value }` at some point. This will fix the focusing bug.

Closes #138


aabounegm (approved):
Mention the fix in the changelog and then merge

PR #144, opened 2020-10-29
Experimental/custom elements

aabounegm:
Since it was not clear how to use the compiled JS version of the components, the purpose of this PR is to change it to use the easier Web Components API.

To do:
- [ ] Figure out how to properly listen to events (related: https://github.com/sveltejs/svelte/issues/3119, https://v2.svelte.dev/guide#firing-events)
- [ ] Investigate the usage of camelCase in prop name
- [ ] Figure out how to pass more complicate data types (instead of just strings) as props. If not possible, document how to use `$$set`.


attractions/button/button.svelte
```diff
@@ -1,3 +1,7 @@
+<!-- TODO: Decide whether to write out all tag names or leave them as null
```
illright, line 1: Yeah, I think it's best to leave them to be specified by users. That way:
1. We don't have to document new things
2. The end user code gets a little more readable since it's evident where the custom elements are coming from

attractions/rollup.config.js
```diff
@@ -1,20 +1,41 @@
...
+      { file: pkg.module, 'format': 'es' },
+      { file: pkg.main, 'format': 'umd', name },
+    ],
+    plugins: [
+      svelte(),
+      resolve(),
+    ],
+  },
+  {
+    // This will likely replace the other configuration above
```
illright, line 24: Not sure if it's a good idea to replace the other configuration with this one. The `customElement` option alters the constructor in a way that might only be compatible with browsers that support custom elements. Perhaps we could somehow distribute two versions of Attractions, one that's suitable for custom elements and a regular one.

attractions/button/button.svelte
illright: And then maybe we could distribute a simple JS file that registers all the components with an opinionated name, such as `a-button` as in this example.

demo/index.html
```diff
@@ -0,0 +1,40 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Web Components demo</title>
+
+  <!-- ultimately should be something like https://unpkg.com/attractions@2.2.0/dist/index.js -->
+  <script src="../attractions/dist/bundle.js"></script>
+</head>
+
+<body>
+  <a-card>
+    <a-button filled="true" danger="true">
```
illright, line 14: is it even possible to use any props that are not boolean or string? I'm pretty sure HTML treats all props as strings

demo/index.html
```diff
@@ -0,0 +1,40 @@
...
+      Alert
+    </a-button>
+    <!-- camelCase props seem to not work (HTML is case-insensitive) -->
+    <a-switch slotLeft="true">
```
illright, line 18: maybe we should reconsider the way we name our boolean and string props to be compatible with this

demo/index.html
```diff
@@ -0,0 +1,40 @@
...
+    // Or using the normal constructor
+    const { Button } = attractions;
+    new Button({
+      target: document.getElementById('container'),
+      props: {
+        filled: true,
+      },
+    });
+
+    // TODO: figure out how to listen to events
```
illright, line 37: You mean the events inside the shadow DOM?

illright:
Damn, I should have read the PR description a little more carefully) oh well, let's keep those comments as discussion threads

attractions/rollup.config.js
aabounegm: I think the [support](https://caniuse.com/custom-elementsv1) is relatively fine. 94.05% of all users is not bad at all, so I wouldn't worry about older browsers.
Also, the other API is way more difficult, in addition to that it does not include the CSS (with `customElements`, they get inlined in the JS), removing any convenience that would come from using custom components (without Svelte) in general.

attractions/button/button.svelte
aabounegm: Actually both way are somewhat incompatible. To remove the names, we would have to specify `<svelte:options tag={null} />`, and then to register the name we use `customElements.register('a-button', attractions.Button)`. I'm not against the idea of distributing 2 versions, but it would be additional writing we need to do, and the 2nd version is only useful for people using another framework (Vue. for instance).
I would imagine that including one script file from a CDN and using ready components right away would be the most common use-case (for the plain JS users) as it is the fastest for prototyping and doesn't need much additional work. The other approach will require polluting the global window (with the variable `attractions`) and users would need to write a lot of duplicated code to register all components they need

demo/index.html
aabounegm: Right, but a string is still truthy, so we should be ok here as long as we don't compare with `=== true` or similar. About other data types, that still needs investigation

attractions/rollup.config.js
illright: Well, Safari is not supported, which is sad, but then again, as long as it is still possible to properly use the library with Svelte, this thing won't be a critical problem

demo/index.html
aabounegm: Or not rename, but add an "alias", perhaps

demo/index.html
aabounegm: It would be convenient if we can listen to them from outside

attractions/button/button.svelte
illright: I genuinely don't think there will be a lot of components. To me it seems like you'll use custom elements when you just want one or two components from this library, if you want to actually build an interface, you employ Svelte itself. So this code to register the elements won't be much

aabounegm: Yes, but if you are using the library from a CDN without Svelte or some other framework, you probably need a very quick prototype and don't really want to spend time registering all components you need, wondering why some aren't working just to find out you forgot to register some of them. It also requires remembering the API and knowing where to find the components to be registered (the global `attractions` object), in addition to polluting the global namespace. You probably wouldn't even want to write a lot of JavaScript.
The main point is that it does no harm to register them, and anyone using a different framework or simply doesn't like the names can rename them

illright: okay, fine, let's register them with the `a-` prefix

demo/index.html
illright: is that even possible?

aabounegm: I mean another prop that would act as a fallback. If `slotLeft` is `false` (or maybe `null` or some other default signifying its absence), check the value of `slotleft`. But that's kinda hacky and I don't like it.
Do we even have that many important props with camelCase names? I mean, people can still use `.$$set`

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+- All components are now registered as `customElements` as well, meaning that they can be used with Svelte (or any framework). This does not affect any existing usage. More information in [the docs](https://illright.github.io/attractions/docs/customElements).
```
illright, line 10: > meaning that they can be used with Svelte

that was possible all along, perhaps you meant "meaning that they can now be used with any framework, not just Svelte"

docs/src/routes/docs/customElements.svx
```diff
@@ -0,0 +1,74 @@
+---
+name: Custom Elements
```
illright, line 2: why the camel case in the name? we usually name files with dashes

README.md
```diff
@@ -38,6 +38,21 @@ autoPreprocess({
 })
 ```
 
+---
+
+Alteratively, the library can be used from a CDN, such as unpkg, and then the components will be registered as custom elements. This is especially useful for quick prototypes that do not need all the features provided.
```
illright, line 43: *Alternatively
Also perhaps put `unpkg` in backticks?

illright (changes requested):

CHANGELOG.md
aabounegm: I actually meant "with**out**"

docs/src/routes/docs/customElements.svx
aabounegm: yeah, I didn't look beyond the routes for reference)

README.md
aabounegm: unpkg is not code to put it in backticks. maybe a link can be more useful, but should it be to [unpkg.com/attractions](https://unpkg.com/attractions) or just [unpkg.com](https://unpkg.com)?

illright: It's not code, but neither is it a word. Let's have it be a link to unpkg.com/attractions

illright (approved):

PR #145, opened 2020-11-01
Allow peer dependency of PostCSS v8.x  as well

aabounegm:


illright (approved):
Wonder if it's sensible to just leave it blindly greater or should we limit it with like "or" operators or smth..

aabounegm:
What would be a reasonable limit in your opinion? Every 8.x release?

illright:
Yeah, major versions 7 and 8

illright (approved):

PR #150, opened 2020-11-14
Close on keypress and time highlighting

RikuVan:
- The dropdown container will close when escape or return keys are pressed
- The time-picker has highlights current time and allows for custom labels
Screenshot 2020-11-14 at 10 12 18

### DIFF
--- a/attractions/time-picker/time-picker.svelte
+++ b/attractions/time-picker/time-picker.svelte
@@ -130,7 +130,9 @@
         <Button noRipple on:click={() => focus = false}>close the time picker</Button>
       </div>
       <Label>
-        <slot name="hours-label">Hours</slot>
+        <slot name="hours-label">
+          <Label>Hours</Label>
+        </slot>
       </Label>
       <div class="column">
         {#each hourValues as hour}
@@ -142,7 +144,9 @@
         {/each}
       </div>
       <Label>
-        <slot name="minutes-label">Minutes</slot>
+        <slot name="minutes-label">
+          <Label>Minutes</Label>
+        </slot>
       </Label>
       <div class="column">
         {#each minuteValues as mins}
@@ -153,7 +157,9 @@
       </div>
       {#if seconds}
         <Label>
-          <slot name="seconds-label">Seconds</slot>
+          <slot name="seconds-label">
+            <Label>Seconds</Label>
+          </slot>
         </Label>
         <div class="column">
           {#each minuteValues as secs}
@@ -183,7 +189,9 @@
         </div>
       {/if}
       <Button on:click={setToNow}>
-        <Clock />
+        <slot name="now-icon">
+          <Clock />
+        </slot>
         <slot name="now-label">now</slot>
       </Button>
     </Dropdown>
--- a/attractions/time-picker/time-picker.svelte
+++ b/attractions/time-picker/time-picker.svelte
@@ -21,6 +21,10 @@
   export let value = null;
   export let format = '%H:%M';
   export let amPmTabName = 'am-pm';
+  export let hoursLabel = 'Hours';
+  export let minutesLabel = 'Minutes';
+  export let secondsLabel = 'Seconds';
+  export let nowLabel = 'Now';
   $: readableFormat = (
     format
       .replace('%H', 'HH')
@@ -101,6 +105,14 @@
     }
   }
 
+  function matchesCurrentHour(hour, selected) {
+    if (!value) {
+      return false;
+    }
+    const currentHour = f12hours ? ((selected.getHours() + 11) % 12) + 1 : selected.getHours();
+    return hour === currentHour;
+  }
+
   const dispatch = createEventDispatcher();
 </script>
 
@@ -121,24 +133,26 @@
       <div class="shown-on-focus">
         <Button noRipple on:click={() => focus = false}>close the time picker</Button>
       </div>
-      <Label>Hours</Label>
+      <Label>{hoursLabel}</Label>
       <div class="column">
-        {#each hourValues as value}
-        <Button on:click={() => setHours(value + 12 * (f12hours && currentAmPm === 'PM' ^ value === 12))}>
-          {value.toString().padStart(2, '0')}
-        </Button>
+        {#each hourValues as hour}
+          <Button
+            on:click={() => setHours(hour + 12 * (f12hours && (currentAmPm === 'PM') ^ (value === 12)))}
+            filled={matchesCurrentHour(hour, value)}>
+            {hour.toString().padStart(2, '0')}
+          </Button>
         {/each}
       </div>
-      <Label>Minutes</Label>
+       <Label>{minutesLabel}</Label>
       <div class="column">
-        {#each minuteValues as value}
-        <Button on:click={() => setMinutes(value)}>
-          {value.toString().padStart(2, '0')}
-        </Button>
+        {#each minuteValues as mins}
+          <Button on:click={() => setMinutes(mins)} filled={value && mins === value.getMinutes()}>
+            {mins.toString().padStart(2, '0')}
+          </Button>
         {/each}
       </div>
       {#if seconds}
-        <Label>Seconds</Label>
+        <Label>{secondsLabel}</Label>
         <div class="column">
           {#each minuteValues as value}
           <Button on:click={() => setSeconds(value)}>
@@ -165,7 +179,7 @@
       {/if}
       <Button on:click={setToNow}>
         <Clock />
-        now
+        {nowLabel}
       </Button>
     </Dropdown>
   </DropdownShell>
--- a/attractions/time-picker/time-picker.svelte
+++ b/attractions/time-picker/time-picker.svelte
@@ -129,38 +129,36 @@
       <div class="shown-on-focus">
         <Button noRipple on:click={() => focus = false}>close the time picker</Button>
       </div>
-      <Label>
         <slot name="hours-label">
           <Label>Hours</Label>
         </slot>
-      </Label>
       <div class="column">
         {#each hourValues as hour}
           <Button
             on:click={() => setHours(hour + 12 * (f12hours && (currentAmPm === 'PM') ^ (value === 12)))}
-            selected={matchesCurrentHour(hour, value)}>
+            selected={matchesCurrentHour(hour, value)}
+          >
             {hour.toString().padStart(2, '0')}
           </Button>
         {/each}
       </div>
-      <Label>
-        <slot name="minutes-label">
-          <Label>Minutes</Label>
-        </slot>
-      </Label>
+      <slot name="minutes-label">
+        <Label>Minutes</Label>
+      </slot>
       <div class="column">
         {#each minuteValues as mins}
-          <Button on:click={() => setMinutes(mins)} selected={value && mins === value.getMinutes()}>
+          <Button 
+            on:click={() => setMinutes(mins)}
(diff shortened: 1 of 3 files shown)


attractions/time-picker/time-picker.svelte
```diff
@@ -21,6 +21,10 @@
   export let value = null;
   export let format = '%H:%M';
   export let amPmTabName = 'am-pm';
+  export let hoursLabel = 'Hours';
```
illright, line 24: Wondering if these things should be slots. What do you think?

attractions/time-picker/time-picker.svelte
```diff
@@ -121,24 +133,26 @@
...
       </div>
-      <Label>Minutes</Label>
+       <Label>{minutesLabel}</Label>
       <div class="column">
-        {#each minuteValues as value}
-        <Button on:click={() => setMinutes(value)}>
-          {value.toString().padStart(2, '0')}
-        </Button>
+        {#each minuteValues as mins}
+          <Button on:click={() => setMinutes(mins)} filled={value && mins === value.getMinutes()}>
```
illright, line 149: Not really digging this way of emphasis. Filled buttons have a semantic value, I'm afraid they cannot simply be used for emphasis like this.
We've been thinking about this emphasis thing too, haven't come up with anything reasonable yet. Maybe use the Dot component to mark the current time?

illright (changes requested):

illright:
Looks pretty solid apart from those comments, thanks! Don't mind the failing build, it's something we need to fix.

attractions/time-picker/time-picker.svelte
RikuVan: Well this solution meets my needs but I have no objection to using slots.

attractions/time-picker/time-picker.svelte
RikuVan: I get your point. I can try something else. I suppose most date/time pickers have a 'selected' style. But I can see what the dot would look like.

attractions/time-picker/time-picker.svelte
illright: Because I believe that in some other cases we use named slots for such customization + it yields more freedom

RikuVan: yeah, I get  that. Did you think to include the label as well in the <slot> or just the text?

attractions/time-picker/time-picker.svelte
RikuVan: I just noticed there is a selected prop on the Button. Would this be appropriate?  The Dot looks a bit odd to my eye.

RikuVan:
I changed the texts to slots and also added them to the docs. About the styling for the currently selected items, I used "selected" at this point but if this is supposed to be reserved for something different in your opinion, I can try something else.
Screenshot 2020-11-15 at 11 06 46

docs/src/routes/docs/components/time-picker.svx
```diff
@@ -78,5 +78,18 @@ Falsy values passed to classes will be disregarded.
...
+## Slots {#slots}
+
+### `hours-label` slot
+
+Defaults to the text "`Hours`".
+
+### `minutes-label` slot
+
+Default to the text "`Minutes`".
+
```
RikuVan, line 90: oops need to add seconds label

attractions/time-picker/time-picker.svelte
illright: Sorry for the delay, I think it's best to include the label element in the slot for proper customization

attractions/time-picker/time-picker.svelte
illright: Well, it kinda clashes with the focus styles, so that might become confusing. And now that I had a second though about it, is it really worth highlighting the current time? Given that there's a "now" button that can automatically select the current time.

aabounegm (dismissed):

RikuVan (commented):
> Well, it kinda clashes with the focus styles, so that might become confusing. And now that I had a second though about it, is it really worth highlighting the current time? Given that there's a "now" button that can automatically select the current time.

The highlighting is not the 'current time' (e.g what pressing now would get you) but the selected time. There was some problem in the app I am using this in terms of users noticing whether a time had been selected or not. Having the user selection highlighted in some ways eliminates this confusion (it is parallel to the use case of date picker where you want the date highlighted). But if don't want this, that is ok, I can make a custom version for my app and leave this part out.

illright:
> The highlighting is not the 'current time' (e.g what pressing now would get you) but the selected time.

Oh, okay, I suppose, we had a misunderstanding then. Then I think `selected` would be appropriate

illright (commented):
So one last final thing would be to move the `<Label>` into slots and I think it should be good to merge

RikuVan:
I placed the `<Label />` inside the slots. With regard to the button for selecting 'now' it would have been possible to have one slot for both the icon and label but I think this leads to extra friction when you only want to change the text (my case) since you would then have to import the <Clock /> icon to include it. Having a separate slot for the icon allows you to change or omit it if needed. So now you have do the following.I thought to add this to the docs as well but I suppose you need to publish a new version before adding examples, since the components come from the doc apps package.json.

```svelte
<TimePicker format="%H %M %S">
    <label for="hours" class="label" slot="hours-label">Tunti</label>
    <label for="minutes" class="label" slot="minutes-label">Minuutti</label>
    <label for="seconds" class="label" slot="seconds-label">Sekuntia</label>
    <span slot="now-icon"><TimeIcon /></span>
    <div slot="now-label">nyt</div>
  </TimePicker>
```

aabounegm:
> but I suppose you need to publish a new version before adding examples

Actually, we are using [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/), which allows us to separate the packages but use whatever is currently in the project tree for the dependency. In other words, it does not download Attractions from npm for the docs, but uses the folder next to it. This allows working on both in parallel without having to publish a new version first.

aabounegm:
Also, can you please pull the latest changes from master? I updated the CI and want to see if the issue is fixed

attractions/time-picker/time-picker.svelte
```diff
@@ -121,29 +129,46 @@
       <div class="shown-on-focus">
         <Button noRipple on:click={() => focus = false}>close the time picker</Button>
       </div>
-      <Label>Hours</Label>
+      <Label>
```
illright, line 132: But now there's a duplicate `<Label>` element

attractions/time-picker/time-picker.svelte
```diff
@@ -121,29 +129,46 @@
...
+      </Label>
       <div class="column">
-        {#each hourValues as value}
-        <Button on:click={() => setHours(value + 12 * (f12hours && currentAmPm === 'PM' ^ value === 12))}>
-          {value.toString().padStart(2, '0')}
-        </Button>
+        {#each hourValues as hour}
+          <Button
+            on:click={() => setHours(hour + 12 * (f12hours && (currentAmPm === 'PM') ^ (value === 12)))}
+            selected={matchesCurrentHour(hour, value)}>
```
illright, line 141: Also I haven't exactly mentioned it earlier (and Prettier is still not set up to guide you in this), but we prefer the closing angle bracket on a separate line:
```svelte
<Button
  props
>
```

illright (changes requested):

illright:
> btw way would be nice if the component took a for prop.

Could you elaborate on this please?

attractions/time-picker/time-picker.svelte
RikuVan: Oh that was :shit:. Fixed. @aabounegm the hooks work fine now--don't need to skip the linting.

RikuVan: Honestly, I did development outside of the project and copy-pasted the code (which is why I had the nested Label) because I had the issues with my local prettier changing the code and the broken linting. How are you guys developing the components?

attractions/time-picker/time-picker.svelte
RikuVan: ok, will fix. yeah I have been using prettier for more than two years in all my work projects. Sort of forgot what it is like to develop js/ts without it on the frontend (backend I still format by hand unfortunately).

attractions/time-picker/time-picker.svelte
illright: Well, I develop in Atom (I don't know if you're aware but Atom's community is falling apart and the editor itself is barely kept alive, let alone feature-rich and up-to-date) so I don't have any sort of Prettier intervening with my code. @aabounegm uses VS code, maybe he can provide some more insight here.

RikuVan:
> > btw way would be nice if the component took a for prop.
> 
> Could you elaborate on this please?

I was missing the word Label in the sentence which is what I was talking about -- sorry for the ambiguity. So the Label component does not take a 'for' attribute (https://www.w3schools.com/tags/att_label_for.asp). This is not at all related to this pull request. But I guess you guys are aware that you are missing many props for attributes that would make  semantic and accessible html. Like "for" on label, aria attributes and so on. I used this library for a very small use case but in my "day job" I could never use it for this reason.

attractions/time-picker/time-picker.svelte
RikuVan: I didn't mean the editor, I meant the project itself. Are you developing the components in the same repo locally, or pulling it out into a different sandbox repo for development?

aabounegm:
Actually, any additional props passed to almost any component will be applied on the underlying element, as mentioned in the [docs](https://illright.github.io/attractions/docs/components/typography). However, I just realized that the `<Label>` component is not actually a `<label>` element under the hood, which is definitely misleading and should be fixed. That is in addition to the issue you linked

RikuVan:
> Actually, any additional props passed to almost any component will be applied on the underlying element, as mentioned in the [docs](https://illright.github.io/attractions/docs/components/typography). However, I just realized that the `<Label>` component is not actually a `<label>` element under the hood, which is definitely misleading and should be fixed. That is in addition to the issue you linked

Ah indeed: ` {...$$restProps}`

attractions/time-picker/time-picker.svelte
aabounegm: I don't quite get the question, but we develop on this repo directly

RikuVan: So maybe you are actually testing out components in the docs app as you are developing them? I am just wondering how you render and test the component as you are working on it, without pulling it out of the components folder and putting it in some small app of your own.

RikuVan: I personally just quickly created a Snowpack app and put the components there so I could get a better dev environment with hot reloading to see how the component works.

illright (approved):

attractions/time-picker/time-picker.svelte
aabounegm: Yes. Initially, we had a small draft page where we just played around with the components to test them, but we threw that out once we had proper docs and just used the docs for "testing", and I said before, the docs do not depend on the version published on npm due to using yarn workspaces.
_If curious, you can take a look here: https://github.com/illright/attractions/commit/1b4ac4c3e5ec6a83e11553702a1d81aabd649eea_

RikuVan:
Did you need something more from me on this one?

illright:
Nope, merging. Thanks!

RikuVan:
Thanks to you guys.

PR #163, opened 2020-11-21
Remove svelte:options tag

### LINKED ISSUE #162: Svelte plugin 'Tag' warning.
Updated to the latest attractions version and now I am getting a warning fo each component:
```(!) Plugin svelte: The 'tag' option is used when generating a custom element. Did you forget the 'customElement: true' compile option?```

I do not need this option. Do you know if there is a way to suppress the warning? Maybe something should be added to the documentation about this.

### PR DESCRIPTION
**Pull request by aabounegm**
To prevent warnings from the Svelte compiler when using the uncompiled version of the library, these changes will add the `<svelte:options tag="..." />` via a preprocessor so that it will be done on the fly (automatically using the file name) and only when needed.

Additionally, I added terser to minify the output files.

Resolves #162

### DIFF
--- a/attractions/rollup.config.js
+++ b/attractions/rollup.config.js
@@ -1,5 +1,6 @@
 import path from 'path';
 import svelte from 'rollup-plugin-svelte';
+import { terser } from 'rollup-plugin-terser';
 import resolve from '@rollup/plugin-node-resolve';
 import autoPreprocess from 'svelte-preprocess';
 import pkg from './package.json';
@@ -42,6 +43,9 @@ export default [
     plugins: [
       svelte(),
       resolve(),
+      terser({
+        module: true,
+      }),
     ],
   },
   {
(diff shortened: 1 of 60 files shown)

### CODE REVIEW

File: attractions/rollup.config.js [hunk 1]
```diff
@@ -31,11 +58,17 @@ export default [
     plugins: [
       svelte({
         customElement: true,
-        preprocess: autoPreprocess({
-          scss: { includePaths: ['./'] },
-        }),
+        preprocess: [
+          prependTagOption(icons),
```
illright, line 62: Shouldn't this be commented out for the time being?

aabounegm: Why? Do you not want the custom elements build at all?
This is what fixes the warnings. There should no longer be any warnings, and the custom elements still build as they did

illright: I don't get how this fixes the warnings. Just because this code is added artificially and therefore not checked by the checker?

illright (approved):

attractions/rollup.config.js
aabounegm: Yes. It is added only when we are building, not when the code is imported from somewhere else

PR #166, opened 2020-12-02
Hotifx/ci vulnerability

aabounegm:


aabounegm:
According to [this discussion](https://github.community/t/on-workflow-run-does-not-work-for-me/128833), "To trigger a workflow runs on the `workflow_run` event, the workflow file should exist on the master or default branch", which means we need to merge it first to see if it works (and to remove that `pull_request_target` one)

.github/workflows/pr_docs.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Publish docs
+        uses: peaceiris/actions-gh-pages@v3.7.3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs
+          destination_dir: ${{ env.PR_NUMBER }}
+          keep_files: true
+
+      - name: Post URL on PR
+        uses: jungwinter/comment@v1
```
illright, line 37: Why a different workflow?

aabounegm: The `unsplash/comment-on-pr` one worked only with the `pull_request` event and didn't support manually giving it the PR number

illright (approved):

PR #168, opened 2020-12-07
Set up Prettier

### LINKED ISSUE #151: Add prettier config
Would make it easier to do pull requests if the prettier settings were checked into the repo, e.g. .prettierrc or in the package.json

illright:
I must say, I don't agree with all of the decisions that Prettier makes, but considering that it's such a ubiquitous tool, I see why we would wanna have it in our project.

Closes #151


aabounegm:
Looks like we need to reconfigure eslint and stylelint to be compatible with Prettier

CHANGELOG.md
```diff
@@ -36,22 +41,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/) and this
...
 
-
 ## [2.1.0] - 2020-08-21
 
 First stable release with proper documentation.
 Previous, undocumented, releases can be found in [the releases section](https://github.com/illright/attractions/releases).
 
-
-[Unreleased]: https://github.com/illright/attractions/compare/v2.2.3...HEAD
+[unreleased]: https://github.com/illright/attractions/compare/v2.2.3...HEAD
```
aabounegm, line 59: Is there a way to force this to be uppercased? I would like for the heading to say "Unreleased"

aabounegm (approved):

CHANGELOG.md
illright: It shouldn't matter, should it? I thought these anchors were case-insensitive

aabounegm: Oh, ok then. I suspected that but didn't check

illright: Well, me neither

PR #173, opened 2020-12-08
Fix the autocomplete dropdown

### LINKED ISSUE #172: The Autocomplete broke down
We changed the DropdownShell component to dispatch `change` events on both opening and closing. The Autocomplete relied on the previous behaviour, that the `change` events are only fired on close, therefore, it was preventing the option dropdown from appearing

illright:
Closes #172


aabounegm (approved):
btw, was this change to the DropdownShell documented in the changelog?

illright:
No, but it was a bug. It's just us who relied on the buggy functionality, other people should have gotten the hint that `change` means any change, not just to `false`. Therefore, I don't think that was worth documenting

aabounegm (approved):

PR #174, opened 2020-12-08
Update peer dependencies to remove warnings

illright:


aabounegm (approved):
I didn't run, as long as you checked it doesn't break anything

PR #175, opened 2020-12-08
Auto-generate an ID if one is not given

### LINKED ISSUE #167: Warning: Found 2 elements with non-unique id #null
TextFields and other components that have the `id` as a prop default to that ID being `null` for some reason, which is the reason of warnings in the console like:
```
Found 2 elements with non-unique id #null
```

aabounegm:
Closes #167
The warning in the console should be harmless, but this will still prevent it. It has the downside of assigning random IDs to elements even when needed, but the upside of linking the `<input>` with its `<label>` even if the user didn't explicitly provide an id (which should add some convenience).

### DIFF
--- a/attractions/form-field/form-field.svelte
+++ b/attractions/form-field/form-field.svelte
@@ -10,6 +10,10 @@
   export let help = null;
   export let id = null;
 
+  if (id == null) {
+    id = `form-field-${Math.random().toString().substring(2, 8)}`;
+  }
+
   export let required = false;
   export let optional = false;
   export let errors = [];
--- a/attractions/pagination/pagination.scss
+++ b/attractions/pagination/pagination.scss
@@ -11,7 +11,7 @@
     > .page {
       width: 2.5em;
       height: 2.5em;
-      margin: 0 .2em;
+      margin: 0 0.2em;
       justify-content: center;
       padding: 0;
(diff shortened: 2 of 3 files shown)


illright:
No, I don't like this. Not only is this just curing the symptom (the `null` valued props shouldn't appear at all, this is not just about the `id` prop), but it is also introducing very shifty behaviour. IDs should be given for a reason, not just to suppress some warnings.

aabounegm:
So do you recommend just waiting until Sapper fixes the problem (we first need to open an issue)? I don't have much faith in that since it seems that their efforts are being shifted from Sapper to SvelteKit nowadays.
I understand that this just a workaround, but our `<label>`s still need to be linked to their `<inputs>`s. Alternatively, we can put the input inside the label and depend on the `id` being in `$$restProps` if needed.

illright:
Yes, I'd rather investigate the issue and let the Sapper team know. Maybe even a PR if we can.
As for linking inputs and labels, yes, this is an accessibility issue and I am currently sketching ways to make Attractions more accessible (locally for now). So don't worry, we'll get this working

aabounegm:
Alright then, I'll close this for now. We might resort back to this solution if they do not fix the issue and your sketches don't work out

PR #178, opened 2020-12-09
Use a download action that supports cross-workflow

aabounegm:
Since the official actions/download-artifact only works within the same workflow
This Action should be able to download the artifact from a different workflow


illright (approved):
I'll trust you on this one

aabounegm:
I don't really trust me, but merging is the only way of finding out if it works)

PR #179, opened 2020-12-11
Cleanup

illright:
Bunch of chores. Finally got around to checking how everything runs, writing a proper `package.json`, satisfying package conventions, all sorts of stuff like that


attractions/package.json
```diff
@@ -2,12 +2,22 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/illright/attractions.git",
+    "directory": "attractions"
+  },
   "svelte": "index.js",
   "module": "dist/index.mjs",
   "main": "dist/index.js",
+  "unpkg": "dist/custom-elements.js",
+  "sideEffects": false,
```
aabounegm, line 17: I think the `custom-elements.js` counts as a side effect, right?

docs/rollup.config.js
```diff
@@ -3,11 +3,12 @@ import replace from '@rollup/plugin-replace';
 import commonjs from '@rollup/plugin-commonjs';
 import svelte from 'rollup-plugin-svelte';
 import babel from '@rollup/plugin-babel';
-import alias from '@rollup/plugin-alias';
 import { terser } from 'rollup-plugin-terser';
 import config from 'sapper/config/rollup.js';
-import autoPreprocess from 'svelte-preprocess';
 import pkg from './package.json';
+
```
aabounegm, line 9: What is this separation for?

docs/src/utils/segment-to-name.js
```diff
@@ -0,0 +1,8 @@
+// Return the word with only the first letter uppercased
```
aabounegm, line 1: I would suggest using JSDoc

docs/rollup.config.js
illright: The imports above are from the template, it's just easier to compare with the template to make sure we're still up-to-date on the best practices

docs/src/utils/segment-to-name.js
illright: Yeah, this is something. I've been looking around the internet today and found [`ibm/sveld`](https://github.com/IBM/sveld) which can generate TS type definitions from JSDoc. We may consider it

attractions/package.json
illright: Hm. Well, that's true, but then again, `custom-elements.js` is not meant to be imported... But then again, `sideEffects` accepts an array of files that have side effects. Yeah, I think we should add it

aabounegm: It is meant to be imported when not using Svelte or a library that is somehow compatible with Svelte components.

docs/src/utils/segment-to-name.js
aabounegm: Cool! It's less than a month old, but I say we go for it

attractions/package.json
illright: Not imported, simply included, since it's an IIFE

aabounegm: But it still exposes a global variable, `attractions`, on which you can find all components and utils

aabounegm (approved):

PR #185, opened 2020-12-20
Escape the regexp in the autocomplete match highlighting

### LINKED ISSUE #184: Autocomplete-option fails when query includes regex pattern
For example, the autocomplete fails on the input `+` because js regex interprets that as 'repeat the last character'.

To reproduce, query `+++` on the documentation page https://illright.github.io/attractions/docs/components/autocomplete

is this intended? If so, the documentation could include that detail. Or maybe this behavior could be exposed as a boolean flag.

illright:
Closes #184

### DIFF
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -1,9 +1,10 @@
 <script>
   import { createEventDispatcher } from 'svelte';
+  import escapeRegExp from '../utils/escape-regexp.js';
 
   export let option;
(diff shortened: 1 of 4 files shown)


aabounegm (dismissed):
Could you maybe also add it to the docs?

illright:
I have a feeling this is not exactly what you had in mind, but I don't see a reason or a way to document this in the `AutocompleteOption`

aabounegm (approved):
It actually is exactly what I had in mind

PR #186, opened 2020-12-20
Fix the autocomplete options not showing up after selection

### LINKED ISSUE #183: Autocomplete does not display options on focus, after a selected item is removed
Hi, I really enjoy using attractions. I am using Autocomplete as a form of dropdown select. In most cases, my application will populate the Autocomplete with just 1-5 selection options, and require 0 characters typed before displaying those options. 

I expect the autocomplete input to show these selection options each time the input is focused.
- And indeed, on the first time the input component is focused, the selection options are displayed. 
- However, if you remove a selected item with X and then click back on the input, the selection options are not displayed until you type again.

My use of the component looks like this

      <Autocomplete 
        {getOptions}
        bind:selection
        minSearchLength=0
        maxOptions=1 />

illright:
Closes #183

### DIFF
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -30,7 +30,7 @@
   let inputElement = null;
 
   function updateOptionGenerator(query) {
-    if (query != null && query.length >= minSearchLength) {
+    if (query.length >= minSearchLength) {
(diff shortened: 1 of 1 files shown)


aabounegm (approved):
Didn't check it as I'm on my phone, but I can't check it soon, so just merge if you are sure

PR #187, opened 2020-12-21
Slot forwarding

illright:


docs/src/routes/docs/components/autocomplete.svx
```diff
@@ -93,6 +93,30 @@ Falsy values passed to classes will be disregarded.
 | **`change`** | `{ value }` | Fired when the selection changes. The `value` contains the selection array. |
 
 
+## Slots {#slots}
+
```
aabounegm, line 97: How about you just mention that the same slots in `autocomplete-field` can be passed here, just to not duplicate any changes to the slots documentation?

attractions/package.json
```diff
@@ -43,13 +45,14 @@
...
     "stylelint": "^13.6.1",
-    "svelte": "^3.24.0",
+    "svelte": "^3.29.1",
     "svelte-preprocess": "^4.5.2",
     "typescript": "^4.0.5"
   },
   "peerDependencies": {
     "postcss": ">=7 <9",
     "sass": "^1.30.0",
+    "svelte": "^3.29.1",
```
aabounegm, line 55: Should this be mentioned in the changelog?

aabounegm (commented):
Don't forget to update the changelog, and also reference the relevant issue here (or reference this PR there)

attractions/package.json
illright: Yes, I was even thinking if this is a major change or not. I suppose it shouldn't be, because keeping your dependencies up-to-date is always a good idea

aabounegm: No, this is definitely not a breaking change. It is just that we documented it

aabounegm (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented here.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- Autocomplete now supports the same slots as AutocompleteField.
```
sallaben, line 11: Nice to see this got implemented! Any particular reason you prefer slot forwarding over the alternate methods of passing component params/writing a custom autocomplete?

illright: Not to say we prefer it this way, we just learnt that slot forwarding is possible and figured it didn't hurt to forward them where it was appropriate. As for writing a custom autocomplete, there are still many reasons to do so, which is why we will continue to provide subcomponents to mix-n-match at one's will.

PR #188, opened 2020-12-21
Fix the overflowing text on the file tiles

### LINKED ISSUE #125: FileDropzone can't handle long unbreakable file names
image
When there are no spaces in the filename, the box is exceeded rather that forcefully wrapping or ellipsis'ing the text

illright:
Closes #125

### DIFF
--- a/attractions/file-input/file-tile.scss
+++ b/attractions/file-input/file-tile.scss
@@ -23,6 +23,8 @@
 
   .filename {
     align-self: end;
+    overflow: hidden;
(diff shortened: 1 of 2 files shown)


aabounegm (dismissed):
It would be great if you also add a `title` with the text content when it overflows, just to still be able to access that information

aabounegm (approved):
Would have been great if there was a way to only add the title if the text actually overflows, but it seems like there is no easy way of achieving this without JavaScript.

PR #189, opened 2020-12-21
Fix the radio buttons (again) being incorrectly styled

illright:


aabounegm (approved):
I really gotta fix that CI

PR #190, opened 2020-12-21
Generate a proper set of favicon images

illright:


aabounegm (approved):
Seems like more work than necessary. Do you want the docs to be installable as a PWA?

illright:
I just used a generator website. It's not necessarily for PWAs, even just for bookmarks

PR #201, opened 2020-12-28
Fix the shadow previews not showing on page navigation

illright:
More like work around, I still don't know what is causing this, perhaps I should file a bug to Sapper

### DIFF
--- a/docs/src/components/docs/shadow-preview.svelte
+++ b/docs/src/components/docs/shadow-preview.svelte
@@ -12,9 +12,7 @@
 </script>
 
 <Popover position={PopoverPositions.LEFT}>
-  <div class="preview" bind:this={preview}>
-    hover me
-  </div>
+  <div class="preview" bind:this={preview}>hover me</div>
   <div slot="popover-content" class="value-popover">
     {value.replace(/\), /g, '),\n')}
   </div>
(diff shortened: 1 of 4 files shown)


docs/src/components/docs/shadow-preview.svelte
```diff
@@ -1,12 +1,18 @@
...
+  let preview;
+
+  onMount(function setBoxShadow() {
+    preview.style.boxShadow = value.slice(0, -1);
+  });
 </script>
 
 <Popover position={PopoverPositions.LEFT}>
-  <div class="preview" style="box-shadow: {value};" />
+  <div class="preview" bind:this={preview}>hover me</div>
```
aabounegm, line 15: Just wanted to point out that this is grammatically incorrect. You hover the mouse **over** the `div`, but you don't hover the `div` itself.
Not a big deal, though

aabounegm (approved):

docs/src/components/docs/shadow-preview.svelte
illright: so this should say "hover over me"?

aabounegm: If you care about grammar, yes. If you think it is not really important in this situation, then never mind.

PR #202, opened 2020-12-28
Add a simple scroll to top button

illright:
Currently it's just chilling at the bottom of the page, later on we'll make it pop up, get the focus right for accessibility, etc. All of that after I think about redesigning the docs

### DIFF
--- a/docs/src/routes/docs/_layout.svelte
+++ b/docs/src/routes/docs/_layout.svelte
@@ -1,10 +1,16 @@
 <script>
+  import { Button } from 'attractions';
+  import { HomeIcon, ArrowUpIcon } from 'svelte-feather-icons';
   import Header from 'src/containers/docs/header.svelte';
   import MobileNavigation from 'src/containers/docs/mobile-navigation.svelte';
   import DesktopNavigation from 'src/containers/docs/desktop-navigation.svelte';
-  import { HomeIcon } from 'svelte-feather-icons';
 
   export let segment;
+
+  function scrollToTop() {
+    window.scrollTo(0, 0);
+    // TODO: manage focus
+  }
 
   const places = [
     {
@@ -149,6 +155,10 @@
   <DesktopNavigation {places} segment={segment || ''} />
   <article>
     <slot />
+    <Button on:click={scrollToTop}>
+      <ArrowUpIcon size="24" class="mr" />
+      scroll to top
+    </Button>
   </article>
 </main>
 
--- a/docs/src/routes/docs/_layout.svelte
+++ b/docs/src/routes/docs/_layout.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { onMount } from 'svelte';
   import { Button } from 'attractions';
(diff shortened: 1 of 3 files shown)


docs/src/routes/docs/_layout.svelte
```diff
@@ -149,6 +155,10 @@
   <DesktopNavigation {places} segment={segment || ''} />
   <article>
     <slot />
+    <Button on:click={scrollToTop}>
```
aabounegm, line 158: I think it would look better `filled` and centered.
Also, how about hiding it when the page is too short to have a scrollbar (such as the [docs homepage](https://illright.github.io/attractions/202/docs)? This can be done with something like:
```js
let scrollbarVisible;
onMount(() => {
    scrollbarVisible = window.innerHeight < document.body.scrollHeight;
});
```

docs/src/routes/docs/_layout.svelte
```diff
@@ -1,11 +1,17 @@
...
   import Header from 'src/containers/docs/header.svelte';
   import MobileNavigation from 'src/containers/docs/mobile-navigation.svelte';
   import DesktopNavigation from 'src/containers/docs/desktop-navigation.svelte';
-  import { HomeIcon } from 'svelte-feather-icons';
 
   export let segment;
 
+  function scrollToTop() {
+    window.scrollTo(0, 0);
+    // TODO: manage focus
```
aabounegm, line 12: care to elaborate on this?

docs/src/routes/docs/_layout.svelte
illright: I know, there are lots of things we could do to make it better, but the point is, I want to rethink the design of the docs entirely and then implement all sorts of nice stuff. For now I just threw it in just to be there

docs/src/routes/docs/_layout.svelte
illright: It should reset the focus on the page to the first focusable element, otherwise keyboard users are unable to use this properly

docs/src/routes/docs/_layout.svelte
aabounegm: I get it, but honestly I think that they are great as they currently are; don't overthink it. In any way, I don't think it's much work adding a prop and centering it (and even the 2 extra lines of code)

docs/src/routes/docs/_layout.svelte
```diff
@@ -1,11 +1,29 @@
...
+  let scrollbarVisible;
+
+  function updateScrollbarVisibility() {
+    if (typeof window === 'undefined') {
+      return;
+    }
+    scrollbarVisible = window.innerHeight < document.body.scrollHeight;
+  }
+  onMount(updateScrollbarVisibility);
+  $: updateScrollbarVisibility(segment);
```
aabounegm, line 25: That doesn't seem to be working when I navigate using the buttons instead of the address bar. If I go to [installation](https://illright.github.io/attractions/202/docs/installation) and then click on the home button in the breadcrumb, the button is still visible even though it is not needed

aabounegm (approved):

PR #205, opened 2021-01-10
Version 3

illright:


aabounegm (commented):
LGTM, but why does it give errors in building the docs? I thought Yarn Workspaces will use the library from the file system directly instead of downloading from npm, or is this because of version mismatch?

aabounegm (approved):

PR #206, opened 2021-01-10
Update the Node version for the release builds

illright:
Not sure if that would help with the memory, but the last PR had built successfully for the branch action, so it should be possible with GitHub Actions


illright:
In the meantime, I'll deploy the docs for the last version manually

aabounegm (approved):

PR #209, opened 2021-01-12
Make the release workflow withstand redundant publishes

illright:
We had a moment where the release workflow failed due to the docs building process running out of memory, however, the publishing job went through alright. The whole workflow was marked as failing, so I decided to rerun it. But since our release job fails when a published version is being published again, we are now eternally stuck with a failing badge on our README for the version 3.0.0, too bad.

This will change up the release workflow in such a way that it hopefully won't fail because of repeated attempts to publish the package. I also rewrote the `check-version.js` script because the output it gives is really messed up and it confused me into thinking something broke down.


aabounegm:
This makes us prone to the mistakes where we forget to update the version in package.json and create a new release without checking the output of the Action. This was a one time problem (not sure how it even happened), but that is curing the symptom not the cause.
IMO, the correct way is to add a dependency between the jobs so that the publishing will not begin before the docs are deployed using the [`jobs.<job>.needs`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) syntax

illright:
I'm not sure how correct this actually is. Those things aren't actually interdependent, and it'd be strange to hold back a working release because GitHub Actions is under the weather. But I agree it does expose us to such mistakes. But then again, how much of a problem is it, really? We have code review, and whenever I publish a release, I look forward to a confirmation email from npm, so I'd notice if something went wrong

aabounegm:
@illright I understand that they aren't interdependent, but I dislike the solution of ignoring redundant versions.
When we used to have a develop branch, it was easier to protect against this kind of mistakes (just another workflow running on PRs to master only to check the package version). We removed that check when we let go of that branch.
Anyway, I think that:

- protecting against this kind of mistakes is more important than being "technically correct" about job independence
- I don't expect the same error to happen again (or often, at least). I think we can just go by without doing anything to protect against it

illright:
So you wanna close this?

aabounegm:
@illright Let's leave it till the next release or something so we see if we face the same problem again, but in principle, yes.

PR #210, opened 2021-01-12
Add a workflow to manually deploy the docs

illright:
This is my attempt at writing an action for manual deployment. It'd be great to also pass `force_orphan` to the value opposite of `keep_files` but since those are strings, not booleans, idk how to invert its value.

### DIFF
--- a/.github/workflows/manual_deploy.yml
+++ b/.github/workflows/manual_deploy.yml
@@ -1,0 +1,40 @@
+name: Manual deployment of docs
+on:
+  workflow_dispatch:
+    inputs:
+      prefix:
+        description: 'The URL prefix to deploy the docs to'
+        required: false
+      keep_files:
+        description: 'Keep other deployments on the branch (true/false)?'
+        required: false
+        default: 'true'
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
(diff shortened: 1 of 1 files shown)


.github/workflows/manual_deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+          yarn install
+          SAPPER_APP_BASEPATH=attractions/${{ github.event.inputs.prefix }} yarn export --basepath attractions/${{ github.event.inputs.prefix }}
+
+      - name: Publish docs
+        uses: peaceiris/actions-gh-pages@v3.7.3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs
+          destination_dir: ${{ github.event.inputs.prefix }}
+          keep_files: ${{ github.event.inputs.keep_files }}
```
aabounegm, line 40: Are you sure this will work, or will it always be truthy (since the string `'false'` is truthy)?

aabounegm (approved):
Not sure why we need this, but ok

.github/workflows/manual_deploy.yml
illright: No idea, truthfully. I have not a single clue how one would manipulate data among the actions. Just blind faith

PR #211, opened 2021-01-13
Typings

aabounegm:
This is to track progress on the typings support.
- [x] Add [`sveld`](https://github.com/IBM/sveld) to the build process
- [x] Fix all incorrectly detected slot types (happens only for slots with props)
- [x] Document all props (description and correct `@type`)
- [x] Document all dispatched events
- [x] Ensure that all `$$restProps` are detected correctly (or add them all for consistency)
- [x] Check if any component [`@extends`](https://github.com/IBM/sveld#extends) another
- [x] Document the utils as well with JSDoc and add it to the type generation

### DIFF
--- a/attractions/accordion/accordion-section.svelte
+++ b/attractions/accordion/accordion-section.svelte
@@ -1,4 +1,7 @@
 <script>
+  /**
+   * @slot {{ toggle: () => void }} handle
+   */
   import Button from '../button/button.svelte';
   import { createEventDispatcher } from 'svelte';
   import classes from '../utils/classes.js';
--- a/attractions/accordion/accordion-section.svelte
+++ b/attractions/accordion/accordion-section.svelte
@@ -9,7 +9,16 @@
   let _class = null;
   export { _class as class };
 
+  /**
+   * The label text to use on the button that toggles the section.
+   * Not used if the handle slot is provided.
+   * @type {string}
+   */
   export let label = null;
+  /**
+   * The state of the section: opened or closed.
+   * @type {boolean}
+   */
   export let open = false;
 
   const selfControl = {
--- a/attractions/accordion/accordion-section.svelte
+++ b/attractions/accordion/accordion-section.svelte
@@ -1,6 +1,8 @@
 <script>
   /**
    * @slot {{ toggle: () => void }} handle
+   * @event {{ close: () => void; toggle: () => void }} panel-open
+   * @event {{ close: () => void; toggle: () => void }} panel-close
    */
   import Button from '../button/button.svelte';
   import { createEventDispatcher } from 'svelte';
--- a/attractions/accordion/accordion.svelte
+++ b/attractions/accordion/accordion.svelte
@@ -7,6 +7,10 @@
 
   let _class = null;
   export { _class as class };
+  /**
+   * Setting this to true allows multiple panels to be open at the same time.
+   * @type {boolean}
+   */
   export let multiple = false;
 
   let currentlyOpenPanel = null;
--- a/attractions/accordion/accordion.svelte
+++ b/attractions/accordion/accordion.svelte
@@ -1,7 +1,7 @@
 <script>
   /**
    * @typedef {import('./accordion-section').default} AccordionSection
-   * @slot {{ closeOtherPanels: ({detail: AccordionSection }) => void }}
+   * @slot {{ closeOtherPanels: (e: CustomEvent<AccordionSection>) => void }}
    */
   import classes from '../utils/classes.js';
 
--- a/attractions/accordion/accordion.svelte
+++ b/attractions/accordion/accordion.svelte
@@ -1,4 +1,8 @@
 <script>
+  /**
+   * @typedef {import('./accordion-section').default} AccordionSection
+   * @slot {{ closeOtherPanels: ({detail: AccordionSection }) => void }}
+   */
   import classes from '../utils/classes.js';
 
   let _class = null;
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -1,4 +1,7 @@
 <script>
+  /**
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   */
   import { createEventDispatcher } from 'svelte';
   import DropdownShell from '../dropdown/dropdown-shell.svelte';
   import Dropdown from '../dropdown/dropdown.svelte';
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -1,6 +1,8 @@
 <script>
   /**
    * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @typedef {typeof import('./autocomplete-option').Option} Option
+   * @typedef {(q: string) => Generator<Promise<Option[]>, never, never>} OptionsGetter
    */
   import { createEventDispatcher } from 'svelte';
   import DropdownShell from '../dropdown/dropdown-shell.svelte';
@@ -16,13 +18,49 @@
   let _class = null;
   export { _class as class };
 
+  /**
+   * An async generator of suggestions.
+   * Receives the input from the text field and is expected to yield promises that resolve to arrays of objects
+   * @type {OptionsGetter}
+   */
   export let getOptions;
+  /**
+   * The current selection as an array of objects.
+   * Can be used to set the selection programmatically.
+   * @type {Option[]}
+   */
   export let selection = [];
+  /**
+   * The minimum amount of characters to call `getOptions`
+   * @type {number}
+   */
   export let minSearchLength = 3;
+  /**
+   * The maximum amount of options than can be selected.
+   * @type {number}
+   */
   export let maxOptions = Infinity;
+  /**
+   * The current value of the text field. Can be used to control the query programmatically.
+   * @type {string}
+   */
   export let searchQuery = '';
+  /**
+   * The component used to render a suggestion in the list.
+   * Receives two props: the option object ({ name, details }) and the current query.
+   * Expected to dispatch click events on selection
+   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: MouseEvent}, {}>}
+   */
   export let optionComponent = AutocompleteOption;
+  /**
+   * Whether to disable the field.
+   * @type {boolean}
+   */
   export let disabled = false;
+  /**
+   * Allows to programmatically control whether the suggestions dropdown is shown.
+   * @type {boolean}
+   */
   export let focus = false;
 
   let moreOptions = false;
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -1,8 +1,9 @@
 <script>
   /**
-   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
    * @typedef {typeof import('./autocomplete-option').Option} Option
    * @typedef {(q: string) => Generator<Promise<Option[]>, never, never>} OptionsGetter
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @event {{ value: Option[] }} change
    */
   import { createEventDispatcher } from 'svelte';
   import DropdownShell from '../dropdown/dropdown-shell.svelte';
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -2,7 +2,7 @@
   /**
    * @typedef {typeof import('./autocomplete-option').Option} Option
    * @typedef {(q: string) => Generator<Promise<Option[]>, never, never>} OptionsGetter
-   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} more-options
    * @event {{ value: Option[] }} change
    */
   import { createEventDispatcher } from 'svelte';
@@ -50,7 +50,7 @@
    * The component used to render a suggestion in the list.
    * Receives two props: the option object ({ name, details }) and the current query.
    * Expected to dispatch click events on selection
-   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: MouseEvent}, {}>}
+   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: Event}, {}>}
    */
   export let optionComponent = AutocompleteOption;
   /**
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -21,18 +21,18 @@
 
   /**
    * An async generator of suggestions.
-   * Receives the input from the text field and is expected to yield promises that resolve to arrays of objects
+   * Receives the input from the text field and is expected to yield promises that resolve to arrays of `Option` objects.
    * @type {OptionsGetter}
    */
   export let getOptions;
   /**
-   * The current selection as an array of objects.
+   * The current selection as an array of `Option` objects.
    * Can be used to set the selection programmatically.
    * @type {Option[]}
    */
   export let selection = [];
   /**
-   * The minimum amount of characters to call `getOptions`
+   * The minimum length the search query must be to call `getOptions`.
    * @type {number}
    */
   export let minSearchLength = 3;
@@ -48,8 +48,8 @@
   export let searchQuery = '';
   /**
    * The component used to render a suggestion in the list.
-   * Receives two props: the option object ({ name, details }) and the current query.
-   * Expected to dispatch click events on selection
+   * Receives two props: `option` – the `Option` object ({ name, details }) and `query` – the current query.
+   * Expected to dispatch click events on selection.
    * @type {SvelteComponentTyped<{option: Option; query: string}, {click: Event}, {}>}
    */
   export let optionComponent = AutocompleteOption;
--- a/attractions/autocomplete/autocomplete-field.svelte
+++ b/attractions/autocomplete/autocomplete-field.svelte
@@ -2,7 +2,7 @@
   /**
    * @typedef {typeof import('./autocomplete-option').Option} Option
    * @typedef {(q: string) => Generator<Promise<Option[]>, never, never>} OptionsGetter
-   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} more-options
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{ nativeEvent: MouseEvent }>) => void }} more-options
    * @event {{ value: Option[] }} change
    */
   import { createEventDispatcher } from 'svelte';
@@ -50,7 +50,7 @@
    * The component used to render a suggestion in the list.
    * Receives two props: `option` – the `Option` object ({ name, details }) and `query` – the current query.
    * Expected to dispatch click events on selection.
-   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: Event}, {}>}
+   * @type {SvelteComponentTyped<{ option: Option; query: string }, { click: Event }, {}>}
    */
   export let optionComponent = AutocompleteOption;
   /**
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -1,6 +1,6 @@
 <script>
   /**
-   * @typedef {{name: string; details?: string}} Option
+   * @typedef {{ name: string; details?: string }} Option
    * @event {{ nativeEvent: MouseEvent }} click
    */
   import { createEventDispatcher } from 'svelte';
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -7,7 +7,8 @@
   import escapeRegExp from '../utils/escape-regexp.js';
 
   /**
-   * @type {Option} The option data to render
+   * The option data to render.
+   * @type {Option}
    */
   export let option;
   /**
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -1,6 +1,7 @@
 <script>
   /**
    * @typedef {{name: string; details: string}} Option
+   * @event {{ nativeEvent: MouseEvent }} click
    */
   import { createEventDispatcher } from 'svelte';
   import escapeRegExp from '../utils/escape-regexp.js';
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -1,8 +1,18 @@
 <script>
+  /**
+   * @typedef {{name: string; details: string}} Option
+   */
   import { createEventDispatcher } from 'svelte';
   import escapeRegExp from '../utils/escape-regexp.js';
 
+  /**
+   * @type {Option}
+   */
   export let option;
+  /**
+   * The substring to seek out and highlight among the name and the details.
+   * @type {string}
+   */
   export let query = null;
   $: matchRegex = query ? new RegExp(`(${escapeRegExp(query)})`, 'ig') : null;
 
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -7,7 +7,7 @@
   import escapeRegExp from '../utils/escape-regexp.js';
 
   /**
-   * @type {Option}
+   * @type {Option} The option data to render
    */
   export let option;
   /**
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -1,6 +1,6 @@
 <script>
   /**
-   * @typedef {{name: string; details: string}} Option
+   * @typedef {{name: string; details?: string}} Option
    * @event {{ nativeEvent: MouseEvent }} click
    */
   import { createEventDispatcher } from 'svelte';
--- a/attractions/autocomplete/autocomplete.svelte
+++ b/attractions/autocomplete/autocomplete.svelte
@@ -4,6 +4,7 @@
    * @slot {{ }} more-options
    * @slot {{ }} not-enough-input
    * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @extends {'./autocomplete-field'} AutocompleteFieldProps
    * @typedef {import('./autocomplete-field').OptionsGetter} OptionsGetter
    * @typedef {import('./autocomplete-option').Option} Option
    */
--- a/attractions/autocomplete/autocomplete.svelte
+++ b/attractions/autocomplete/autocomplete.svelte
@@ -36,7 +36,7 @@
    * @type {Option[]}
    */
   export let selection = [];
-  
+
   /**
    * The minimum amount of characters to call `getOptions`
    * @type {number}
--- a/attractions/autocomplete/autocomplete.svelte
+++ b/attractions/autocomplete/autocomplete.svelte
@@ -25,14 +25,14 @@
   export { _class as class };
 
   /**
-   * The current selection as an array of objects.
+   * The current selection as an array of `Option` objects.
    * Can be used to set the selection programmatically.
    * @type {Option[]}
    */
   export let selection = [];
 
   /**
-   * The minimum amount of characters to call `getOptions`
+   * The minimum length the search query must be to call `getOptions`.
    * @type {number}
    */
   export let minSearchLength = 3;
--- a/attractions/autocomplete/autocomplete.svelte
+++ b/attractions/autocomplete/autocomplete.svelte
@@ -3,9 +3,9 @@
    * @typedef {import('./autocomplete-field').OptionsGetter} OptionsGetter
    * @typedef {import('./autocomplete-option').Option} Option
    * @slot {{ }} loading-options
-   * @slot {{ }} more-options
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} more-options
    * @slot {{ }} not-enough-input
-   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @slot {{ }} too-many-options
    * @event {{ value: Option[] }} change
    * @extends {'./autocomplete-field'} AutocompleteFieldProps
    */
--- a/attractions/autocomplete/autocomplete.svelte
+++ b/attractions/autocomplete/autocomplete.svelte
@@ -4,6 +4,8 @@
    * @slot {{ }} more-options
    * @slot {{ }} not-enough-input
    * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} too-many-options
+   * @typedef {import('./autocomplete-field').OptionsGetter} OptionsGetter
+   * @typedef {import('./autocomplete-option').Option} Option
    */
   import { createEventDispatcher } from 'svelte';
   import Button from '../button/button.svelte';
@@ -20,12 +22,45 @@
   let _class = null;
   export { _class as class };
 
+  /**
+   * An async generator of suggestions.
+   * Receives the input from the text field and is expected to yield promises that resolve to arrays of objects
+   * @type {OptionsGetter}
+   */
   export let getOptions;
+  /**
+   * The current selection as an array of objects.
+   * Can be used to set the selection programmatically.
+   * @type {Option[]}
+   */
   export let selection = [];
+  
+  /**
+   * The minimum amount of characters to call `getOptions`
+   * @type {number}
+   */
   export let minSearchLength = 3;
+  /**
+   * The maximum amount of options than can be selected.
+   * @type {number}
+   */
   export let maxOptions = Infinity;
+  /**
+   * The current value of the text field. Can be used to control the query programmatically.
+   * @type {string}
+   */
   export let searchQuery = '';
+  /**
+   * Whether to disable the field.
+   * @type {boolean}
+   */
   export let disabled = false;
+  /**
+   * The component used to render a suggestion in the list.
+   * Receives two props: the option object ({ name, details }) and the current query.
+   * Expected to dispatch click events on selection
+   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: MouseEvent}, {}>}
+   */
   export let optionComponent = AutocompleteOption;
 
   let focus = false;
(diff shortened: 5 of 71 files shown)


attractions/accordion/accordion-section.svelte
```diff
@@ -1,12 +1,24 @@
 <script>
+  /**
+   * @slot {{ toggle: () => void }} handle
```
illright, line 3: What about the default slot?

attractions/accordion/accordion.svelte
```diff
@@ -1,8 +1,16 @@
 <script>
+  /**
+   * @typedef {import('./accordion-section').default} AccordionSection
+   * @slot {{ closeOtherPanels: ({detail: AccordionSection }) => void }}
```
illright, line 4: Spaces around braces

attractions/accordion/accordion-section.svelte
aabounegm: They are detected correctly automatically. The only problem is with slots with props.

attractions/autocomplete/autocomplete-option.svelte
```diff
@@ -1,8 +1,19 @@
 <script>
+  /**
+   * @typedef {{name: string; details: string}} Option
```
illright, line 3: The `details` field is optional

attractions/autocomplete/autocomplete-field.svelte
```diff
@@ -13,13 +19,49 @@
...
+  /**
+   * The current value of the text field. Can be used to control the query programmatically.
+   * @type {string}
+   */
   export let searchQuery = '';
+  /**
+   * The component used to render a suggestion in the list.
+   * Receives two props: the option object ({ name, details }) and the current query.
+   * Expected to dispatch click events on selection
+   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: MouseEvent}, {}>}
```
illright, line 53: Not exactly a `MouseEvent`, it can also be a custom event with any details, so maybe it's worth relaxing the condition here

attractions/autocomplete/autocomplete-option.svelte
```diff
@@ -1,8 +1,19 @@
...
+   * @event {{ nativeEvent: MouseEvent }} click
+   */
   import { createEventDispatcher } from 'svelte';
   import escapeRegExp from '../utils/escape-regexp.js';
 
+  /**
+   * @type {Option}
```
illright, line 10: Why no docstring here? It is pretty self-explanatory, but I imagine it would be confusing to users to leave this without a docstring

illright: Maybe we could use something like `The option data to render.`

attractions/autocomplete/autocomplete.svelte
```diff
@@ -14,12 +24,45 @@
   let _class = null;
   export { _class as class };
 
+  /**
+   * An async generator of suggestions.
+   * Receives the input from the text field and is expected to yield promises that resolve to arrays of objects
+   * @type {OptionsGetter}
+   */
   export let getOptions;
```
illright, line 32: I suppose we can remove these props that are only passed down to `AutocompleteField` to decrease redundancy

attractions/autocomplete/autocomplete.svelte
```diff
@@ -1,4 +1,14 @@
 <script>
+  /**
+   * @typedef {import('./autocomplete-field').OptionsGetter} OptionsGetter
+   * @typedef {import('./autocomplete-option').Option} Option
+   * @slot {{ }} loading-options
+   * @slot {{ }} more-options
```
illright, line 6: This slot also has props, the same as in the `autocomplete-field`

attractions/button/button.svelte
```diff
@@ -1,26 +1,85 @@
...
   export let href = null;
+  /**
+   * Disables Sapper prefetching for link buttons (those with an href prop).
+   */
   export let noPrefetch = false;
 
-  export let events = null;
+  /**
+   * List of handlers for the [Events](https://illright.github.io/attractions/docs/utilities) action
+   * @type {Array<{name: string; handler: (e:Event) => void }>}
```
illright, line 80:
```suggestion
   * @type {Array<{ name: string; handler: (e: Event) => void }>}
```

attractions/checkbox/checkbox-group.svelte
```diff
@@ -1,19 +1,55 @@
...
+   * Designates the checkboxes as color choices.
+   * Labels are ignored and values are expected to be `"#RRGGBB"`.
+   * @type {boolean}
+   */
   export let color = false;
 
+  /**
+   * An array of items, where only the value field is required.
+   * Must contain at least one value.
+   * @type {Array<{value: string; label: string; checked: boolean; disabled: boolean}>}
```
illright, line 35:
```suggestion
   * @type {Array<{ value: string; label: string; checked: boolean; disabled: boolean }>}
```

attractions/checkbox/checkbox.svelte
```diff
@@ -1,18 +1,58 @@
...
+   */
   export let inputClass = null;
+  /**
+   * A class string to add to the selector box element.
+   * @type {string}
+   */
   export let selectorClass = null;
+  /**
+   * A CSS style string to assign to the selector box element.
+   * Can be used to make the checkbox represent a color (in conjunction with the `getColorPickerStyles` utility).
```
illright, line 22: Should we link to the utility here?

attractions/checkbox/checkbox.svelte
```diff
@@ -1,18 +1,58 @@
...
+  /**
+   * Whether the checkbox is checked or not.
+   * @type {boolean}
+   */
   export let checked = false;
+  /**
+   * The value assigned to the checkbox input. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value) for more information.
+   * @type {string}
+   */
   export let value = null;
```
illright, line 36: Unfortunate inconsistency with the radio button, however, in this case having a single checkbox and not having to pass a value to it could be valuable. But maybe we should cut the tiny convenience and opt for consistency and compliance with standards.

illright:
Allow me to make a little checklist for myself:
- [ ] Prop types
- [ ] Prop comments
- [ ] Slots
- [ ] Events
- [ ] Rest props
- [ ] Whitespace

attractions/snackbar/snackbar.svelte
```diff
@@ -6,14 +6,47 @@
...
+   */
   export let closeCallback;
+  /**
+   * The transition to use to animate the in and out of the snackbar.
+   * @type {() => import('svelte/transition').TransitionConfig}
+   */
   export let transition = fly;
+  /**
+   * The options to use with the given transition.
+   * @type {Record<string, any>}
```
illright, line 48: I don't think we should be imposing any limitations here, I'd say have it be an `object`

attractions/snackbar/snackbar.svelte
```diff
@@ -6,14 +6,47 @@
...
   export let closeOnAction = true;
+  /**
+   * The callback to call to close the snackbar in the `SnackbarContainer`.
+   * Passed internally by the `SnackbarContainer` when using the `showSnackbar` function.
+   * @type {() => void}
+   */
   export let closeCallback;
+  /**
+   * The transition to use to animate the in and out of the snackbar.
+   * @type {() => import('svelte/transition').TransitionConfig}
```
illright, line 43: This typing will forbid passing parameters to the transition, we should use this instead:
```suggestion
   * @type {(...args: any[]) => import('svelte/transition').TransitionConfig}
```

attractions/autocomplete/autocomplete-field.svelte
aabounegm: You mean make it just `Event` or `any` or what exactly?

attractions/autocomplete/autocomplete-option.svelte
aabounegm: Honestly, just because I didn't know what to write, and it seemed self-explanatory.

attractions/autocomplete/autocomplete.svelte
aabounegm: By "these props", what are you referring to, next to `getOptions`? I see `optionComponent` as well, and I guess that's all

attractions/checkbox/checkbox.svelte
aabounegm: If you mean with the `@see` or `@link` tags, they will not make it through sveld. It only passes the text description and tags it understands through. If you mean a normal markdown link, to what should it link? The docs or the implementation?

attractions/checkbox/checkbox.svelte
aabounegm: I don't get it. Do you mean you want the `value` to be required (not have a default value)?

attractions/checkbox/checkbox.svelte
```diff
@@ -1,18 +1,58 @@
...
   export let selectorStyle = null;
 
+  /**
+   * Whether the checkbox is checked or not.
+   * @type {boolean}
+   */
   export let checked = false;
+  /**
+   * The value assigned to the checkbox input. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value) for more information.
+   * @type {string}
```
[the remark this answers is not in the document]
aabounegm: I thought Svelte allowed this only with `<select>`s. Wouldn't the object be stringified to `[object Object]`?

attractions/chip/checkbox-chip.svelte
```diff
@@ -1,20 +1,64 @@
...
   export let chipClass = null;
 
+  /**
+   * Whether the chip is selected or not.
+   * @type {boolean}
+   */
   export let checked = false;
+  /**
+   * The value of this checkbox. Check [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value) for more information.
+   * Included in events unaltered (won't be stringified if not a string).
```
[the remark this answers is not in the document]
aabounegm: I left that note there just because you wrote it, but now I don't see how it is even valid. In the dispatched event, `value` is taken from `e.target.value`; is that not stringified by the browser?

attractions/chip/radio-chip-group.svelte
```diff
@@ -1,14 +1,39 @@
...
 
+  /**
+   * An array of items, where only the value field is required.
+   * Must contain at least one value.
+   * @type {Array<{value: string; label: string; disabled: boolean}>}
+   */
   export let items;
+  /**
+   * The currently selected value of the group.
+   * @type {any}
```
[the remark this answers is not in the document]
aabounegm: because this `value` is only used in `bind:group` which can be any type (right?). Should it be `string` here?

attractions/chip/radio-chip.svelte
```diff
@@ -1,21 +1,64 @@
...
   export let inputClass = null;
+  /**
+   * A class string to add to the underlying Chip element.
+   * @type {string}
+   */
   export let chipClass = null;
 
+  /**
+   * The value of this radio button. Included in events and can be bound to using `bind:group`.
+   * @type {any}
```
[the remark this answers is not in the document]
aabounegm: Same reasoning as above

attractions/dialog/dialog.svelte
```diff
@@ -5,11 +5,31 @@
...
   export let titleClass = null;
 
+  /**
+   * Applies the danger color to the title (including the icon slot if it is stylable with the CSS `color` property).
+   * @type {boolean}
+   */
   export let danger = false;
+  /**
+   * Adds a close button to the dialog and calls this function when it is clicked.
+   * @type {(e: MouseEvent) => void}
```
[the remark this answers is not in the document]
aabounegm: Actually `CustomEvent<{ nativeEvent: MouseEvent }>`

attractions/file-input/file-dropzone.svelte
```diff
@@ -10,11 +14,39 @@
...
+   */
   export let fileComponent = FileTile;
+  /**
+   * Limits the allowed files to particular types. For guidelines on the value of the attribute, consult the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept).
+   * @type {string}
+   */
   export let accept = null;
+  /**
+   * A callback to call for each file that passes the `accept` check before it is added to the `files`.
+   * If it returns a Promise, they will be started for every file, but awaited at the end.
```
illright, line 37:
```suggestion
   * If it returns a Promise, they will be started for every file in parallel and awaited together at the end.
```

attractions/snackbar/snackbar-container.svelte
```diff
@@ -1,9 +1,23 @@
...
+   *   expired: Promise<boolean>,
+   * } }}
+   */
   import { setContext } from 'svelte';
   import Snackbar from './snackbar.svelte';
   import snackbarContextKey from './snackbar-context-key.js';
   import SnackbarPositions from './snackbar-positions.js';
 
+  /**
+   * The position of the snackbar stack inside the window.
```
illright, line 18:
```suggestion
   * The position of the snackbar stack inside the container.
```

attractions/snackbar/snackbar-container.svelte
```diff
@@ -1,9 +1,23 @@
...
+   * @type {typeof import('../../snackbar/snackbar-positions').default}
+   */
   export let position = SnackbarPositions.BOTTOM_LEFT;
   let registeredSnackbars = new Set();
```
illright, line 22: Can't comment on the relevant line, but I guess the `showSnackbar` function also needs to be documented as the DropdownShell's `toggle`.

attractions/star-rating/star-rating.svelte
```diff
@@ -8,13 +11,37 @@
...
+   * @type {number}
+   */
   export let max = 5;
+  /**
+   * Current amount of selected stars.
+   * @type {number}
+   */
   export let value = null;
+  /**
+   * The name to assign to all stars belonging to the same group. Check MDN docs for more information.
```
illright, line 31:
```suggestion
   * The name to assign to all stars belonging to the same group. Check [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefname) for more information.
```

attractions/form-field/form-field.svelte
```diff
@@ -3,15 +3,48 @@
...
+   */
   export let required = false;
+  /**
+   * Marks the form field as optional, explicitly stating the optionality of the field in the text under the field.
+   * @type {boolean}
+   */
   export let optional = false;
+  /**
+   * An array of error messages to display under the field. Falsy elements are not shown.
+   * @type {string[]}
```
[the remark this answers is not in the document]
aabounegm: It is not exactly `false`, but any falsey value. This includes `false`, `null`, `undefined`, `0`, `""`, and `NaN`. There is no type for that in typescript, and it feels too ugly to write out entirely. Either we create an alias `type Falsy = false | null | undefined | 0 | ""` (note that `NaN` cannot be used as a type), or we limit it to just exactly `false`, in which case we need to update a lot of the documentation
Note that using `null` will now throw an error unless `strictNullChecks` is on, which would actually break most of our types.

attractions/rollup.config.js
```diff
@@ -77,6 +80,33 @@ export default [
...
+            ],
+            { shell: true }
+          );
+          // Since writeBundle hook is parallel, we need to wait for sveld to finish first
+          setTimeout(() => {
+            appendFileSync(
+              './types/index.d.ts',
+              'export * as utils from "./utils";\n'
+            );
+          }, 1000);
```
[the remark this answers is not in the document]
aabounegm: Yes. Should I increase it to 2 seconds?
The problem is that sveld uses the `writeBundle` hook, which is the last hook to run, and also runs in parallel instead of sequentially, so I couldn't find a cleaner solution.
Of course, another option is to remove this "plugin" entirely and just add the following 2 lines to the CI, but then types will not be available when building locally. They aren't available anyways on Windows due to some stupid bug with globbing, but I thought why limit you (Linux user)?
```bash
yarn tsc utils/*.js --declaration --allowJs --emitDeclarationOnly --outDir types/utils
echo "export * as utils from './utils';" >> types/index.d.ts
```

attractions/text-field/text-field.svelte
```diff
@@ -1,31 +1,96 @@
...
   }
 
+  /**
+   * The current value of the text field. Converted to a number if `type="number"`.
+   * @type {string | number}
+   */
   export let value = null;
-  export let events = null;
+  /**
+   * List of handlers for the [Events](https://illright.github.io/attractions/docs/utilities) action
```
illright, line 90:
```suggestion
   * List of handlers for the [Events](https://illright.github.io/attractions/docs/utilities) action.
```

attractions/autocomplete/autocomplete-field.svelte
illright: We can make it a `MouseEvent | CustomEvent` with any template argument.

attractions/autocomplete/autocomplete.svelte
illright: Yeah, not too much. But still

attractions/checkbox/checkbox.svelte
illright: Yeah, I mean the normal Markdown link. Link to the docs, like we do with any other utility.

illright: By the way, there's a typo in the docs, the utility is called `getColorPickerStyles`, not `getColorPickerStyle`

attractions/checkbox/checkbox.svelte
illright: Yeah, I suppose making it required would be the best way to go.

attractions/checkbox/checkbox.svelte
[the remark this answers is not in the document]
illright: It would, but it doesn't matter, everything still works, it's just that the DOM representation is meaningless.
But I think we shouldn't rely too much on that, and demand them all to be strings.

attractions/chip/checkbox-chip.svelte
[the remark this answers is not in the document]
illright: No, it isn't. Perhaps the target is actually the component instance, not the DOM element. But still, bad behaviour!

attractions/chip/radio-chip-group.svelte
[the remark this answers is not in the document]
illright: It should be string everywhere.

attractions/form-field/form-field.svelte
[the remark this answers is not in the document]
illright: I know, but the main usecase for this thing is writing `errors = [condition && 'error', another && 'other']`, so it's `false` that we need to allow mostly.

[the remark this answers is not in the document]
illright: Why would `strictNullChecks` break our types?

attractions/autocomplete/autocomplete-field.svelte
aabounegm: Well, it's not like we are using this event anyways, so I just made it `Event` since both of these inherit from it.

attractions/checkbox/checkbox.svelte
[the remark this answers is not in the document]
aabounegm: No, I don't think it is just a representation. I remember facing many warning about having duplicate `id`s (or something else, not quite sure): `[object Object]`. They would probably have the same value. This needs testing

attractions/chip/checkbox-chip.svelte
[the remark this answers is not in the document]
aabounegm: So what do you suggest? Mark the type as `any` or remove the comment?

attractions/checkbox/checkbox.svelte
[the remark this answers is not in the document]
aabounegm: I will leave it as `string` for consistency as you said in another comment

attractions/snackbar/snackbar.svelte
aabounegm: `object` means any non-primitive, which includes arrays and basically anything constructed using `new`. `Record` is the TypeScript way of saying "the object we are familiar with that we can represent as a literal `{ key: value }`"

attractions/autocomplete/autocomplete-option.svelte
```diff
@@ -1,8 +1,19 @@
 <script>
+  /**
+   * @typedef {{name: string; details?: string}} Option
+   * @event {{ nativeEvent: MouseEvent }} click
+   */
   import { createEventDispatcher } from 'svelte';
   import escapeRegExp from '../utils/escape-regexp.js';
 
+  /**
+   * @type {Option} The option data to render
```
illright, line 10: Is that even valid? Looks rather strange, it'd be better to put the description above the type (and the full stop).

attractions/date-picker/date-picker.svelte
```diff
@@ -18,17 +22,63 @@
...
+   * The language tag defining the desired locale (e.g., `en-US`). If left `undefined`, the user's locale will be used.
+   * This will affect the weekdays and the day number representations.
+   * @type {string | undefined}
+   */
   export let locale = undefined;
+  /**
+   * The index of the weekday to start the week at.
+   * 0 is for Sunday and 6 is for Saturday.
+   * Defaults to 1 (Monday)
+   * @type {number}
```
illright, line 56: The same thing here for the constraint to actual values.

attractions/autocomplete/autocomplete-option.svelte
aabounegm: yeah, that's just me being mindless

attractions/form-field/form-field.svelte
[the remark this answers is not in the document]
aabounegm: Because `let x: string = null` is not valid with `strictNullChecks`, which we have almost everywhere.
```js
/** @type {string} */
export let value = null;
```
is basically doing just that.
Although the issue may not be that severe if we don't expect the users to pass in `null` at all and we just use it as a default value.

attractions/snackbar/snackbar-container.svelte
aabounegm: It is documented in the top as a slot prop. `export function ...` syntax is not currently being handled correctly by sveld; it detects it as a normal prop whose default value is that function.

attractions/snackbar/snackbar.svelte
illright: Yeah, but there can be a transition that takes an array as an option. This is entirely up to the transition object, therefore, this constraint might cause errors.

attractions/form-field/form-field.svelte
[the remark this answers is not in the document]
illright: Well, that says something about our typings. Perhaps we should add `null` in many places since it is technically a valid value for a lot of props

attractions/snackbar/snackbar.svelte
aabounegm: And can't there be a transition that takes a number (or any primitive value)? Isn't `any` more appropriate then?

attractions/chip/radio-chip.svelte
```diff
@@ -1,21 +1,64 @@
 <script>
+  /**
+   * @event {{ value: any; nativeEvent: Event }} change
```
illright, line 3: This value should also be replaced to `string`

attractions/chip/radio-chip-group.svelte
```diff
@@ -1,14 +1,39 @@
 <script>
+  /**
+   * @event {{ value: any; nativeEvent: Event }} change
```
illright, line 3: `string` here too

attractions/autocomplete/autocomplete-field.svelte
```diff
@@ -1,4 +1,10 @@
 <script>
+  /**
+   * @typedef {typeof import('./autocomplete-option').Option} Option
+   * @typedef {(q: string) => Generator<Promise<Option[]>, never, never>} OptionsGetter
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} more-options
```
illright, line 5:
```suggestion
   * @slot {{ loadMoreOptions: (click?: CustomEvent<{ nativeEvent: MouseEvent }>) => void }} more-options
```

attractions/autocomplete/autocomplete-field.svelte
```diff
@@ -13,13 +19,49 @@
...
+  /**
+   * The current value of the text field. Can be used to control the query programmatically.
+   * @type {string}
+   */
   export let searchQuery = '';
+  /**
+   * The component used to render a suggestion in the list.
+   * Receives two props: `option` – the `Option` object ({ name, details }) and `query` – the current query.
+   * Expected to dispatch click events on selection.
+   * @type {SvelteComponentTyped<{option: Option; query: string}, {click: Event}, {}>}
```
illright, line 53:
```suggestion
   * @type {SvelteComponentTyped<{ option: Option; query: string }, { click: Event }, {}>}
```

attractions/autocomplete/autocomplete-option.svelte
```diff
@@ -1,8 +1,20 @@
 <script>
+  /**
+   * @typedef {{name: string; details?: string}} Option
```
illright, line 3:
```suggestion
   * @typedef {{ name: string; details?: string }} Option
```

attractions/autocomplete/autocomplete.svelte
```diff
@@ -1,4 +1,14 @@
 <script>
+  /**
+   * @typedef {import('./autocomplete-field').OptionsGetter} OptionsGetter
+   * @typedef {import('./autocomplete-option').Option} Option
+   * @slot {{ }} loading-options
+   * @slot {{ loadMoreOptions: (click?: CustomEvent<{nativeEvent: MouseEvent}>) => void }} more-options
```
illright, line 6:
```suggestion
   * @slot {{ loadMoreOptions: (click?: CustomEvent<{ nativeEvent: MouseEvent }>) => void }} more-options
```

attractions/date-picker/date-picker.svelte
```diff
@@ -1,4 +1,8 @@
 <script>
+  /**
+   * @typedef {{start: Date; end: Date}} DateRange
```
illright, line 3:
```suggestion
   * @typedef {{ start: Date; end: Date }} DateRange
```

attractions/file-input/file-dropzone.svelte
```diff
@@ -10,11 +14,39 @@
...
+   * The user's selection as an array of [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) objects.
+   * Note: this is an actual `Array`, not a `FileList`, as opposed to the native `<input type="file">`.
+   * @type {File[]}
+   */
   export let files = [];
+  /**
+   * The component used to render a selected file.
+   * Receives a single prop: `file`, the [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) object.
+   * Expected to dispatch `delete` events to delete this file from the selection with the `file` prop as the detail.
+   * @type {SvelteComponentTyped<{file: File}, {delete: CustomEvent<File>}>}
```
illright, line 27:
```suggestion
   * @type {SvelteComponentTyped<{ file: File }, { delete: CustomEvent<File> }>}
```

attractions/snackbar/snackbar.svelte
```diff
@@ -6,14 +6,47 @@
...
   export let buttonClass = null;
 
+  /**
+   * The text to show on the snackbar.
+   * @type {string}
+   */
   export let text;
+  /**
+   * The action for the button on the snackbar. If this is null, the button is not rendered. Otherwise it has text as a label and calls callback on click.
+   * @type {{text: string; callback: () => void}}
```
illright, line 27:
```suggestion
   * @type {{ text: string; callback: () => void }}
```

attractions/star-rating/star-rating.svelte
```diff
@@ -8,13 +11,37 @@
...
   export let name;
+  /**
+   * Disables star selection.
+   * @type {boolean}
+   */
   export let disabled = false;
-  export let events = null;
+  /**
+   * List of handlers for the [Events](https://illright.github.io/attractions/docs/utilities) action
+   * @type {Array<{name: string; handler: (e:Event) => void }>}
```
illright, line 42:
```suggestion
   * @type {Array<{ name: string; handler: (e: Event) => void }>}
```

attractions/form-field/form-field.svelte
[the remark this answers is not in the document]
aabounegm: Maybe, but it also doesn't always make much sense to pass `null` for something when you can just leave it. The only problem might be when a user binds a variable whose type is `T|null` and our prop accepts just `T` (with `strictNullChecks` of course), but I don't know how often could this be. I am also not sure if Svelte itself will be strict somehow. I think it is best to either test it or wait until someone opens an issue

attractions/snackbar/snackbar.svelte
illright: True, make it `any` then

attractions/snackbar/snackbar-container.svelte
illright: Yeah, but it's also exposed as a regular prop. In `dropdown-shell.svelte` you added a comment to the exported function, regardless of how sveld supports it, I mean we should add a similar comment here

aabounegm: I did, but it's not like it matters. It still didn't give it the correct type. It might be worth it only for the description (for the time being, until sveld fixes the issue)

attractions/rollup.config.js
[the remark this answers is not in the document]
aabounegm: Scratch that, even 2 seconds is not deterministic enough.
There is a few options here:
1. Add more time
2. Move these 2 lines to the CI
3. Move them to a node script (so that it becomes one line instead of 2 from the CLI)
4. Move sveld to the first build and the custom "plugin" to the second one. Note that it seems that the builds run sequentially, even though the documentation states that they run in parallel

Note that, in conjunction, with the above methods, we can create a `tsconfig.json` to make it work cross-platform and not just on Linux.
What do you think?

attractions/form-field/form-field.svelte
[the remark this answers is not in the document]
aabounegm: I'll mark this as resolved (as the main comment has been addressed), and we can discuss the `strictNullChecks` issue if a related issue arises.

attractions/utils/events.js
```diff
@@ -1,4 +1,9 @@
-/* An action to set up arbitrary event listeners dynamically. */
+/**
+ * An action to set up arbitrary event listeners dynamically.
+ * @param {Element} node
+ * @param {Array<{name: string; handler: EventListenerOrEventListenerObject}>} args The event listeners to be registered
```
illright, line 4:
```suggestion
 * @param {Array<{ name: string; handler: EventListenerOrEventListenerObject }>} args The event listeners to be registered
```

illright: Also not sure about that `EventListenerOrEventListenerObject`, what is that supposed to be? This thing should simply take callbacks, is that it?

attractions/utils/dynamic-transition.js
```diff
@@ -1,5 +1,11 @@
-/* Create a transition that allows specifying a transition programmatically
-   or disable it altogether. */
+/**
+ * Create a transition that allows specifying a transition programmatically
+ *  or disable it altogether.
+ * @typedef {{ delay: number; duration: number; css: () => string}} Transition
```
illright, line 4:
```suggestion
 * @typedef {{ delay: number; duration: number; css: () => string }} Transition
```

attractions/utils/datetime-utils.js
```diff
@@ -170,6 +206,13 @@ export function datesLessEqual(date1, date2) {
   );
 }
 
+/**
+ * Generates a calendar view of a given month.
+ * @param {0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11} month Zero-based numeric value for month. 0 = January
+ * @param {number} year
+ * @param {0 | 1 | 2 | 3 | 4 | 5 | 6} firstWeekday First day of the week. 1 = Monday
+ * @returns {Array<Array<{value: Date; outside: boolean}>>}
```
illright, line 214:
```suggestion
 * @returns {Array<Array<{ value: Date; outside: boolean }>>}
```

attractions/utils/color-picker-styles.js
```diff
@@ -1,5 +1,9 @@
-/* Return the styles needed for displaying a radiobutton or a checkbox
-   with the given color. */
+/**
+ * Generates the styles needed for displaying a radiobutton or a checkbox
+ *  with the given color
```
illright, line 3:
```suggestion
 *  with the given color.
```

attractions/utils/classes.js
```diff
@@ -1,9 +1,19 @@
...
+ * @param {string} classString The string to be trimmed
+ * @returns {string} Clean class string
+ */
 export function stripClassWhitespace(classString) {
   return classString.replace(consecutiveSpace, ' ').trimRight();
 }
 
+/**
+ * Filters out falsy classes
+ * @param  {...string|false} args The classes to be filtered
```
illright, line 14:
```suggestion
 * @param {...Array<string | false>} args The classes to be filtered
```
Shouldn't it be like that?

attractions/utils/classes.js
```diff
@@ -1,9 +1,19 @@
...
+ * Removes extra whitespace in class strings
+ * @param {string} classString The string to be trimmed
+ * @returns {string} Clean class string
+ */
 export function stripClassWhitespace(classString) {
   return classString.replace(consecutiveSpace, ' ').trimRight();
 }
 
+/**
+ * Filters out falsy classes
```
illright, line 13:
```suggestion
 * Filters out falsy classes.
```

attractions/utils/classes.js
```diff
@@ -1,9 +1,19 @@
 const consecutiveSpace = / {2,}/;
 
+/**
+ * Removes extra whitespace in class strings
```
illright, line 4:
```suggestion
 * Removes extra whitespace in class strings.
```

attractions/utils/call-on-sight.js
```diff
@@ -1,6 +1,11 @@
-/* Action that allows calling a function when an element appears in the viewport.
-   Uses the IntersectionObserver API, which might not be available in older browsers like IE11,
-   If it is not available, the action will be silently ignored. */
+/**
+ * Action that allows calling a function when an element appears in the viewport.
+ * Uses the IntersectionObserver API, which might not be available in older browsers like IE11.
+ * If it is not available, the action will be silently ignored.
+ * @param {Element} node The HTML element to be observed
+ * @param {{ callback: (...args: any[]) => void; args: any[]}} options Callback and its args
```
illright, line 6:
```suggestion
 * @param {{ callback: (...args: any[]) => void; args: any[] }} options Callback and its args
```

attractions/utils/classes.js
aabounegm: No, the `...` already implies it's an array

attractions/utils/events.js
aabounegm: It is defined internally as `type EventListenerOrEventListenerObject = EventListener | EventListenerObject;`
`EventListener` is the regular callback you are used to, and `EventListenerObject` is an object that has the handler called `handleEvent`. Read more in the [MDN docs for the event listener callback](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback).
They type of `node.addEventListener` is the following
```ts
Element.addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void
```
so I just used the same type for the listener we register.

attractions/rollup.config.js
```diff
@@ -77,6 +79,17 @@ export default [
...
       resolve(),
+      sveld({
+        typesOptions: {
+          preamble: 'export * as utils from "./utils";\n',
+        },
+      }),
+      {
+        name: 'utilsTypings',
+        writeBundle() {
+          spawnSync('tsc', { shell: true });
```
illright, line 90: Isn't there a nice TypeScript plugin for Rollup?

illright (approved):

attractions/rollup.config.js
aabounegm: Yes, but it will affect the bundle. It cannot simply do a side effect and leave the rest of the build as it is. In fact, they explicitly state that it will ignore the `emitDeclarationOnly` (among other options) and set it to `false`

PR #213, opened 2021-01-15
Various actions improvements

illright:
These are on a per-commit basis. Notably, this will set up Yarn caching (finally), replace some commenting action with [GitHub Script](https://github.com/actions/github-script) (that thing is really cool) and pin the versions of all third-party actions for security.

Pardon my eagerness to play with fire, that's my way of learning GitHub Actions :p


aabounegm (approved):
LGTM

PR #214, opened 2021-01-16
Add a CodeQL workflow

illright:
Triggered on PRs to `main` or manually


aabounegm (approved):
I didn't read about CodeQL, so I'll assume what you wrote is correct

PR #218, opened 2021-01-17
Prevent GitHub Actions from posting a docs URL if one already exists

illright:


aabounegm (approved):
I forgot that `workflow_run` Actions run only from the master branch. I guess no other way to test it then

PR #220, opened 2021-01-17
Allow pull request commits to skip CI

illright:


aabounegm:
Any idea why the checks were skipped here? The commit message contains "skip ci", not "[skip ci]"

.github/workflows/pr_lint_build.yml
```diff
@@ -3,6 +3,7 @@ on: pull_request
 
 jobs:
   lint:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
```
aabounegm, line 6: `head_commit` is only available for `push` events. According to [this post](https://github.community/t/github-actions-does-not-respect-skip-ci/17325/7), it should be `toJSON(github.event.commits.*.message)`, though I'm not sure if that would skip for all commits or just this particular sync

illright:
Honestly, I hardly care about this feature, feels like it's more trouble than it's worth

PR #222, opened 2021-01-17
Add a workflow to suggest releases from certain workflows

illright:


.github/workflows/suggest_release.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Post the URL on the PR 🖊
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const package = JSON.parse(require('fs').readFileSync('attractions/package.json'));
```
aabounegm, line 26: Same as in other PR:
 ```suggestion
            const package = require(`${process.env.GITHUB_WORKSPACE}/attractions/package.json`);
```

File: .github/workflows/suggest_release.yml [hunk 2]
```diff
@@ -36,4 +36,12 @@ jobs:
                 issue_number: context.issue.number,
                 labels: ['new release'],
               });
+              await github.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body: |
+                  This pull request will automatically trigger a release on merge!
+                  If it shouldn't, remove the [new release] label.
```
aabounegm, line 45: Won't this "`[new release]`" be detected as markdown?

illright: No, it doesn't have () after it

aabounegm: I thought it will be treated as a reference link, but it doesn't matter much

aabounegm (approved):

PR #224, opened 2021-01-17
Fix the step names and use native Node JSON imports

illright:


.github/workflows/auto_release.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ steps.set-result.outputs.result }}
-          release_name: ${{ steps.set-result.outputs.result }}
-          body: ${{ steps.set-changelog.outputs.result }}
+          tag_name: ${{ steps.fetch-version.outputs.result }}
+          release_name: ${{ steps.fetch-version.outputs.result }}
+          body: ${{ steps.fetch-changelog.outputs.result }}
```
aabounegm, line 51: Don't you mean
```suggestion
          body: ${{ steps.fetch-version.outputs.result }}
```
here as well?

illright: No, body takes the changelog as input

aabounegm: ugh, that step was hidden so I thought it doesn't exist. pardon me

aabounegm (approved):

PR #226, opened 2021-01-20
Add ability for initial selection of a radio group button

EMH333:
If the `value` parameter of a radio-group is set, and there is an item with that same value, that button is initially checked. This is more in line with expected behavior. This doesn't appear to change anything after the initial selection.


illright:
I believe the behaviour you described was working before, this was powered by the line `bind:group={value}`, so I don't think there's a need for this. Please, correct me if I'm wrong

EMH333:
You may very well be right! I missed that initially. I've been swamped these last few days so let me try that and get back to you. I'll close this for now assuming you are right and will reopen if required

PR #227, opened 2021-01-20
Update installation instructions

### LINKED ISSUE #217: The build breaks with Webpack 4
We have a [wildcard export in our `index.js`](https://github.com/illright/attractions/blob/master/attractions/index.js#L74) that Webpack 4 doesn't understand yet. Thus, we need to advise the Webpack 4 users to add an additional Babel transpilation step for this particular file to build the library successfully.

illright:
This will also include a warning notice about Webpack 4.

Closes #217

### DIFF
--- a/docs/src/routes/docs/installation.svx
+++ b/docs/src/routes/docs/installation.svx
@@ -65,11 +65,69 @@ name: Installation
   </div>
 {/if}
 
+{#if yourFramework !== Framework.other && yourBundler === Bundler.webpack}
+  <div class="caution">
+    <Label>Caution</Label>
+
+    Webpack 4.x.x doesn't support the following ES Modules syntax: `export * as X from 'Y'`.
+    Thus, in order to use Attractions with Webpack 4, you need to transpile its index file with Babel.
+
+    Install Babel (if you don't have it already):
+
+    <CopyableCode terminal>
+      {#if yourPackageManager === PackageManager.npm}
+
+```shell
+npm install --save-dev babel-loader @babel/core @babel/preset-env
+```
+
+      {:else if yourPackageManager === PackageManager.yarn}
+
+```shell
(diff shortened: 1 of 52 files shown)


docs/src/routes/docs/installation.svx
```diff
@@ -3,110 +3,380 @@ name: Installation
...
+
+  </CopyableCode>
+
+  {#if yourBundler === Bundler.rollup}
+    {#if yourFramework === Framework.svelte}
+      <CopyableCode filename="rollup.config.js">
+
+```js
+import svelte from 'rollup-plugin-svelte';
+import svelteConfig from './svelte.config.js';
```
aabounegm, line 210: I don't think this works. You cannot mix ES6 modules with CommonJS modules. I know it is suggested in the docs of `svelte-preprocess`, but I never succeeded in actually importing the config.
I usually just leave the `svelte.config.js` file there for the Svelte VS Code extension to pick up on it.

Perhaps it will work with Webpack (as it uses CommonJS all the way), but not with Rollup (as `svelte.config.js` is required by the Svelte Language Server to be in CommonJS).

aabounegm (commented):
I love the work you did on the installation page ✨ 👍

docs/src/routes/docs/installation.svx
aabounegm: Actually, I just tried it again now, and it magically worked. I thought we cannot mix ES6 and CommonJS imports, but it seems like Rollup has been updated or something and it is no longer complaining

aabounegm (approved):

docs/src/routes/docs/installation.svx
illright: No, actually, it wasn't working, for various reasons, but now it's all good.

aabounegm (approved):

PR #235, opened 2021-01-22
Exclude the line containing the version from the release body

illright:
And fix an embarassing case of undefined variables


aabounegm (changes requested):
Can't comment on it, but line 38 should be `script: |`

aabounegm:
And same for 28, actually

aabounegm (approved):

PR #237, opened 2021-01-31
Update dependencies

aabounegm:
In Sapper v0.29, `rel="prefetch"` has changed to `sapper:prefetch` (which makes some sense). However, we have the following line in our `Button` component:
https://github.com/illright/attractions/blob/4c485af6429a51ea86ee35c0e3ea2cfb6f8095b3/attractions/button/button.svelte#L104
Which raises 2 questions:
1. Would changing that line to `sapper:prefetch={noPrefetch ? null : true}` work? (on the premise that Svelte will filter out any `null` attributes). Note that Sapper checks only the existence of that attribute, not its value.
2. Would this be considered a breaking change or is it just something to mention in the changelog (with the assumption that anyone who upgrades dependencies will upgrade all of them at once).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,18 @@ module.exports = {
     browser: true,
     es6: true,
   },
-  extends: ['@sveltejs', 'prettier'],
+  extends: ['prettier'],
+  parserOptions: {
+    ecmaVersion: 2021,
+    sourceType: 'module',
+  },
+  plugins: ['svelte3'],
+  overrides: [
+    {
+      files: ['*.svelte'],
+      processor: 'svelte3/svelte3',
+    },
+  ],
   settings: {
     'svelte3/ignore-styles': attrs => attrs.lang === 'scss',
   },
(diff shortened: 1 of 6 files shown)


aabounegm:
For the second point, I think we can start with just a deprecation (keeping both attributes) until people get a chance to upgrade and then remove the old one. The [keep a changelog](https://keepachangelog.com/en/1.1.0/) format we are using has a `Deprecated` section.

illright:
1. Yep, just tried it, it works
2. Yeah, keeping both would be a good solution

illright (approved):

illright:
However, I don't think we need to write anything in our Changelog. It's not our change, not our breaking change, not our deprecation. People using Sapper will figure it out themselves.

aabounegm:
Not even as an "addition" that our latest version supports Sapper's latest updates?

illright:
Well, we could say that, but just in the release comment, no deprecations or anything

PR #238, opened 2021-01-31
Typings improvement

aabounegm:
The _types_ folder was not being picked up by TypeScript as it is not something special for it (unlike `@types/attractions`, but that's for when the types are unofficial and contributed by the community). Apparently, the `"types"` field of _package.json_ is only used for the main import, i.e.: `import { Button } from "attractions";`, and not when importing submodules, i.e.: `import { classes } from "attractions/utils";`. It refers only to a file (_index.d.ts_). Everything else in the directory not being directly imported by it is ignored.

It was also not possible to add type declarations for a module when its untyped (JS) source code exists next to it. In other words, we cannot write `declare module "attractions/utils" { /* ... */ }` in _index.d.ts_ since the actual _utils_ (source code) folder exists and it doesn't have types.

Therefore, the solution implemented here just puts the type declarations right next to their source code while being gitignored. This way, any js/svelte file has its types right next to it and can be imported directly, not just through the root `"attractions"` module.
This also has the added benefit that the type imports paths are now the same inside the _svelte_ files and don't try to `../..` their way out of the _types_ folder (looking at snackbar/popover positions in particular)

### DIFF
--- a/attractions/importer.js
+++ b/attractions/importer.js
@@ -1,6 +1,6 @@
 /**
- * Creates a custom Sass importer for painless configurations
- * @param {{themeFile?: string; nodeModulesPath?: string}} [options]
+ * Creates a custom Sass importer for painless configurations.
+ * @param {{ themeFile?: string; nodeModulesPath?: string }} [options]
  * @returns {(url: string, prev?: string) => ({ file: string } | null)}
  */
 module.exports = function makeAttractionsImporter(options = {}) {
--- a/attractions/importer.js
+++ b/attractions/importer.js
@@ -1,3 +1,8 @@
+/**
+ * Creates a custom Sass importer for painless configurations
+ * @param {{themeFile?: string; nodeModulesPath?: string}} [options]
+ * @returns {(url: string, prev?: string) => ({ file: string } | null)}
+ */
 module.exports = function makeAttractionsImporter(options = {}) {
   const { themeFile, nodeModulesPath = 'node_modules' } = options;
 
--- a/attractions/text-field/text-field.svelte
+++ b/attractions/text-field/text-field.svelte
@@ -1,7 +1,7 @@
 <script>
   /**
-   * @event {{ value: string | number; nativeEvent: Event }} input
-   * @event {{ value: string | number; nativeEvent: Event }} change
+   * @event {{ value: string | number | null; nativeEvent: Event }} input
+   * @event {{ value: string | number | null; nativeEvent: Event }} change
    * @event {{ nativeEvent: FocusEvent }} focus
    * @event {{ nativeEvent: KeyboardEvent }} keydown
    * @event {{ nativeEvent: FocusEvent }} blur
@@ -63,7 +63,7 @@
   export let label = null;
   /**
    * The error message to show under the text field.
-   * @type {string | null}
+   * @type {string | false | null}
    */
   export let error = null;
   /**
--- a/attractions/text-field/text-field.svelte
+++ b/attractions/text-field/text-field.svelte
@@ -15,17 +15,17 @@
   export { _class as class };
   /**
    * A class string to assign to the `<input>` or `<textarea>` element.
-   * @type {string}
+   * @type {string | false | null}
    */
   export let inputClass = null;
   /**
    * A class string to assign to the `<label>` for the outline text fields.
-   * @type {string}
+   * @type {string | false | null}
    */
   export let labelClass = null;
   /**
    * A class string to add to the error message under the text field.
-   * @type {string}
+   * @type {string | false | null}
    */
   export let errorClass = null;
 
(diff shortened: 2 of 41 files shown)


aabounegm:
Since `--strictNullChecks` (or even `--strict`) is such a nice flag to use (I personally use it in my projects), I think we should rethink which parameters should be publicly documented as nullable and explicitly add ` | null` everywhere applicable. It will also **not** affect those **without** `--strictNullChecks` since TypeScript will just not show that `null` is a possible value (since everything becomes nullable anyways)

attractions/importer.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Creates a custom Sass importer for painless configurations
```
illright, line 2:
```suggestion
 * Creates a custom Sass importer for painless configurations.
```

attractions/importer.js
```diff
@@ -1,3 +1,8 @@
...
+ * @param {{themeFile?: string; nodeModulesPath?: string}} [options]
```
illright, line 3:
```suggestion
 * @param {{ themeFile?: string; nodeModulesPath?: string }} [options]
```

illright (changes requested):

illright (approved):

attractions/utils/range.js
```diff
@@ -1,8 +1,8 @@
 /**
  * Generates a semi-open range.
  * @param {number} start The beginning of the range (included)
- * @param {number} end The end of the range (excluded)
- * @param {number} step The distance between the numbers in the range
+ * @param {number} [end] The end of the range (excluded)
```
illright, line 4: Technically it's `start` that is optional, not `end`

illright (changes requested):

illright:
Maybe we could also use a template type here?

https://github.com/illright/attractions/blob/typings-improvement/attractions/utils/call-on-sight.js#L6

attractions/utils/accepted-file-type.js
```diff
@@ -1,6 +1,6 @@
 /**
  * Validates a File object against an `accept` parameter value.
- * @param {string} references comma-delimited list of reference MIME types (similar to `input`'s `accept` attribute)
+ * @param {?string} references comma-delimited list of reference MIME types (similar to `input`'s `accept` attribute)
```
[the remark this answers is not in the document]
aabounegm: It means just that, `string | null`. Optionality would be `@param {string} [references]`. The thing is that sveld copies the content of the [`@type {}` JSDoc comment](https://jsdoc.app/tags-type.html) exactly as it is, but JSDoc has a syntax slightly different to TypeScript. Of course we can use TypeScript notation everywhere (since it will only be parsed by TS, not JSDoc), but I thought why not just use JSDoc notation? It can help in generating docs later on

attractions/utils/range.js
aabounegm: You cannot have an optional parameter followed by a non-optional one. Again, we are limited by TypeScript's lack of support for overloading in JSDoc.

attractions/utils/accepted-file-type.js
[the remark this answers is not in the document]
illright: Idk, to me it looks very strange, and if we were to generate docs, we'd probably have it powered by TypeScript, but whatever

attractions/utils/call-on-sight.js
```diff
@@ -2,8 +2,9 @@
  * Action that allows calling a function when an element appears in the viewport.
  * Uses the IntersectionObserver API, which might not be available in older browsers like IE11.
  * If it is not available, the action will be silently ignored.
+ * @template T
  * @param {Element} node The HTML element to be observed
- * @param {{ callback: (...args: any[]) => void; args: any[] }} options Callback and its args
+ * @param {{ callback: (...args: T[]) => void; args: T[] }} options Callback and its args
```
illright, line 7: Doesn't that restrict people to arguments of just one type? Can't the template type be an array itself?

aabounegm: It _kind of_ can in TS ([Variadic tuple type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types)), but not in [JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template).

aabounegm: Also, they can always define `T` to be a custom interface and just use one parameter wrapped in a singleton array

attractions/text-field/text-field.svelte
```diff
@@ -58,12 +58,12 @@
...
   /**
    * The label to show above the text field. Only works with `outline` text fields.
-   * @type {string}
+   * @type {string | null}
    */
   export let label = null;
   /**
    * The error message to show under the text field.
-   * @type {string}
+   * @type {string | null}
```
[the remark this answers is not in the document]
illright: Yes, it is

illright (approved):

Issue #243, opened 2021-02-14
[Bug] Date picker: February displays incomplete set of days


sallaben:
This is an interesting one -- I am looking into it myself, but I thought you might want to know as well: February does not display correctly in the attractions datepicker.

Reproduced in the docs: https://illright.github.io/attractions/docs/components/date-picker

![image](https://user-images.githubusercontent.com/7699235/107886022-a966ef80-6eba-11eb-8596-8fa446bd52ca.png)

illright:
Cannot reproduce on desktop/mobile Chrome and on Firefox. Could you tell me what's your browser and system locale?

sallaben:
Mozilla/5.0, Gecko/20100101 Firefox/85.0 -- for what it's worth, I can reproduce on mobile (Safari). Maybe it's a locale based bug?

aabounegm:
I also couldn't reproduce on Firefox 84 or 85.0.2, and my locale is also set to `en-US`. I have no access to Safari to test it there, though.

sallaben:
Very interesting, thanks for the comments guys... I looked deeper into the calendar component and datetime utils, and the issue boils down to the setMonth functionality on JS Date:

```
  const foo = new Date(0);
  foo.setFullYear(2021, 1); // February 2021 (setMonth is used by setFullYear implementation as well)
  console.log(foo);
```

We expect this return a Date object for February 1st, and when I run this code in a UTC node container, I get the expected result: `2021-02-01T00:00:00.000Z`. But when I run the exact same code locally in node (15.8.0) or in my browser, the result is `2021-03-04T00:00:00.000Z`. I'm thinking maybe the way we create the day cursor, using new Date(0), is failing in non UTC environments (I am in Mountain Standard Time).

Here's the gist, from what I understand. When the date is, for example, May 31 2021, and we set the month to February, it becomes February 31 2021 -- which then becomes March 3 2021, since there is no February 31.
- Here are two StackOverflow questions about this topic. [1](https://stackoverflow.com/questions/51581452/wrong-javascript-month-after-2020) [2](https://stackoverflow.com/questions/23395934/javascript-date-bug-february-2014)
- I have identified a fix for this behavior. When using setMonth or setFullYear to change the month value, ensure that we also pass a date value in at the same time.

Here is a PR to fix the behavior in the datetime-utils: #244  

And February now displays normally :-) 
[...]

PR #244, opened 2021-02-14
Bug: fix getCalendar day cursor in short months

sallaben:
- Create day cursor in local timezone

For more discussion see #243

### DIFF
--- a/attractions/utils/datetime-utils.js
+++ b/attractions/utils/datetime-utils.js
@@ -215,14 +215,12 @@ export function datesLessEqual(date1, date2) {
  */
 export function getCalendar(month, year, firstWeekday) {
   const calendar = [];
-  const dayCursor = new Date(0);
-  dayCursor.setFullYear(year, month);
+  const dayCursor = new Date();
 
   // Offset the start of the month to the closest left `firstWeekday`
-  dayCursor.setDate(
-    1 - ((daysInWeek + dayCursor.getDay() - firstWeekday) % daysInWeek)
-  );
-
+  const date = 1 - ((daysInWeek + dayCursor.getDay() - firstWeekday) % daysInWeek);
+  dayCursor.setFullYear(year, month, date);
+  
   do {
     const week = [];
     for (let i = 0; i < daysInWeek; ++i) {
--- a/attractions/utils/datetime-utils.js
+++ b/attractions/utils/datetime-utils.js
@@ -216,11 +216,13 @@ export function datesLessEqual(date1, date2) {
 export function getCalendar(month, year, firstWeekday) {
   const calendar = [];
   const dayCursor = new Date();
+  dayCursor.setFullYear(year, month, 1);
 
   // Offset the start of the month to the closest left `firstWeekday`
-  const date = 1 - (firstWeekday % daysInWeek);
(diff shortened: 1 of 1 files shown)


sallaben:
Thanks, github-actions bot. Verified the change works on the 244 docs as well. 
image

attractions/utils/datetime-utils.js
```diff
@@ -215,12 +215,12 @@ export function datesLessEqual(date1, date2) {
...
   const calendar = [];
-  const dayCursor = new Date(0);
-  dayCursor.setFullYear(year, month);
+  const dayCursor = new Date();
+  dayCursor.setFullYear(year, month, 1);
 
   // Offset the start of the month to the closest left `firstWeekday`
   dayCursor.setDate(
-    1 - ((daysInWeek + dayCursor.getDay() - firstWeekday) % daysInWeek)
+    0 - ((dayCursor.getDay() + firstWeekday) % daysInWeek)
```
aabounegm, line 223: I understand the motive behind removing `daysInWeek`, it does seem redundant with the modulo, but it is actually needed due to the way JavaScript calculates the modulo of negative numbers (which remains negative). Also, why did you change the `1` into `0` and `-` into `+`? It now marks today (15th of February) as Thursday instead of Monday for me (I currently live in UTC+3 timezone, btw).

image

aabounegm (changes requested):
It seems that the same issue affects the weekday labels as well. I suggest changing the `new Date(0)` in line 148 of the same file to `new Date(1970, 0, 1)`

attractions/utils/datetime-utils.js
sallaben: This change was erroneous -- a blind attempt to match the dates to the weekday labels that displayed on my screen. 

Your comment pointing out that my week started on Friday clued me in that the labels on my screen were also being calculated wrongly. You also pointed out that getWeekdays also used new Date(0). Once I updated getWeekdays to use new Date(1970, 0, 1), the labels matched up and I saw that the change to the logic here wasn't warranted.

attractions/utils/datetime-utils.js
```diff
@@ -215,12 +215,12 @@ export function datesLessEqual(date1, date2) {
  */
 export function getCalendar(month, year, firstWeekday) {
   const calendar = [];
-  const dayCursor = new Date();
-  dayCursor.setFullYear(year, month, 1);
+  const dayCursor = new Date(1970, 0, 1);
+  dayCursor.setFullYear(year, month);
```
aabounegm: Couldn't these 2 lines be combined as such:
```suggestion
  const dayCursor = new Date(year, month, 1);
```
?

attractions/utils/datetime-utils.js
aabounegm: Cool! Thanks for **your** help in finding and fixing bugs!

attractions/utils/datetime-utils.js
sallaben: Thanks, I did not know about that equivalency! Committed your suggested change.

aabounegm (approved):
LGTM

illright (approved):

aabounegm:
Thank you so much for your help, @sallaben!

PR #247, opened 2021-02-21
Adds variables to customize buttons & textfield outline

### LINKED ISSUE #245: Allow customizing button font weight
As a user of `attractions`, I need a way to customize the font-weight of buttons, so that I can match my design mockups.

Currently, all `<Button>` instances are set to `font-weight: vars.$bold-font-weight;`.

### LINKED ISSUE #246: Allow customizing outline TextField
As a user of `attractions`, I need a way to customize the outline variant of a `TextField`, so that I can match my design mockups.

Currently, all `<TextField outline>` instances are set to `border-radius: 1.5625em;`. Also, the label itself has a hard-coded `background-color: #fff;`. These are problematic because the rounded nature of the `outline` variant doesn't match any of my designs, and the `background-color: #fff;` doesn't look right if the containing element doesn't have the same background color.

Example:

Screen Shot 2021-02-21 at 12 16 09 PM

damonbauer:
* Introduces a new variable, `$button-font-weight`, to allow changing the `font-weight` of a `<Button>`.
* Introduces a new variable, `$textfield-outline-radius`, to allow changing the `border-radius` of a `<TextField outline>`.
* ~~Introduces a new variable, `$textfield-outline-label-bg`, to allow changing the `background-color` of a `label` in a `<TextField outline>`.~~

* Closes #245 
* Closes #246

### DIFF
--- a/attractions/text-field/text-field.scss
+++ b/attractions/text-field/text-field.scss
@@ -119,7 +119,7 @@
 
     label {
       font-family: vars.$font;
-      background-color: vars.$textfield-outline-label-bg;
+      background-color: vars.$background;
       color: vars.$textfield-border;
       padding: 0 0.25em;
       position: absolute;
--- a/attractions/text-field/text-field.scss
+++ b/attractions/text-field/text-field.scss
@@ -126,6 +126,7 @@
       top: 0;
       left: 1.4em;
       transform: translate(-0.25em, -55%);
+      border-radius: 0.3125em;
     }
 
     &.with-item.left {
--- a/attractions/text-field/text-field.scss
+++ b/attractions/text-field/text-field.scss
@@ -119,14 +119,13 @@
 
     label {
       font-family: vars.$font;
-      background-color: #fff;
+      background-color: vars.$textfield-outline-label-bg;
       color: vars.$textfield-border;
       padding: 0 0.25em;
       position: absolute;
       top: 0;
       left: 1.4em;
       transform: translate(-0.25em, -55%);
-      border-radius: 0.3125em;
     }
 
     &.with-item.left {
@@ -148,7 +147,7 @@
     }
 
     input {
-      border-radius: 1.5625em;
+      border-radius: vars.$textfield-outline-radius;
       border: 1px solid vars.$textfield-border;
       background-color: transparent;
       padding: 0 1.4em;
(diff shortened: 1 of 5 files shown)


attractions/text-field/text-field.scss
```diff
@@ -119,14 +119,13 @@
...
       font-family: vars.$font;
-      background-color: #fff;
+      background-color: vars.$textfield-outline-label-bg;
       color: vars.$textfield-border;
       padding: 0 0.25em;
       position: absolute;
       top: 0;
       left: 1.4em;
       transform: translate(-0.25em, -55%);
-      border-radius: 0.3125em;
```
illright, line 129: Why'd you remove this? The reason why we had it in the first place was, for example, Chrome autocompletion when it gives a bluish tint to the background of the field. That bluish background exposed the rough edges of the label and it wasn't looking too good.

attractions/text-field/text-field.scss
```diff
@@ -119,14 +119,13 @@
 
     label {
       font-family: vars.$font;
-      background-color: #fff;
```
illright, line 122: Oh, did we really have this hardcoded? Oops. We have a variable for the background already though, `$background`, maybe we could use that here? I can't imagine a usecase where this would be different from the main background of the page.

docs/src/routes/docs/components/button.svx
```diff
@@ -133,6 +133,6 @@ The content of the Button. Icons, other components, everything is fair game here
 | **`$shadow-raised`** | The shadow of filled buttons on hover. | <ShadowPreview value={shadowRaised} /> |
 | **`$disabled`** | The text color of disabled buttons. | <ColorPreview value="#888888" /> |
 | **`$disabled-bg`** | The background color of filled disabled buttons. | <ColorPreview value="#D7D7D7" /> |
-| **`$bold-font-weight`** | The font weight of the buttons. | `500` |
+| **`$button-font-weight`** | The font weight of the buttons. | `$bold-font-weight` (`500`) |
```
illright, line 136: I don't think it's worth mentioning that `$button-font-weight` has a value of `$bold-font-weight`. Just `500` should do it

illright (changes requested):

docs/src/routes/docs/components/button.svx
aabounegm: I think it should just be `$bold-font-weight` since it is used as this variable's default. If users modify `$bold-font-weight` but not `$button-font-weight`, it will still change, to the confusion of those who see the default as `500`.

illright: In that case perhaps we should explicitly write out the value of the `$button-font-weight` in the `_variables.scss` instead. I don't like this referencing thing, it looks misleading as if changing one would change the other or something.

attractions/text-field/text-field.scss
damonbauer: Sorry, that was a mistake! Added back.

damonbauer:
@illright @aabounegm Updated!

illright (approved):

aabounegm (approved):

PR #250, opened 2021-03-03
Change the event in the `on-release` workflow

illright:


aabounegm:
This still won't fix it as events dispatched by Actions cannot trigger other Actions. Instead, you can try to use the `workflow_run` trigger:
```yaml
workflow_run:
    workflows:
      - 'Release automatically'
``` 
and don't forget the `if: ${{ github.event.workflow_run.conclusion == 'success' }}`

PR #257, opened 2021-05-03
Fix Snackbar positions type

### LINKED ISSUE #256: Incorrect typings of Snackbar positions
Screenshot from 2021-05-02 10-21-20
Im trying to change position but im not able to.

Error message:
Type 'string' is not assignable to type 'Readonly<{ TOP_LEFT: string; TOP_MIDDLE: string; TOP_RIGHT: string; MIDDLE_RIGHT: string; BOTTOM_RIGHT: string; BOTTOM_MIDDLE: string; BOTTOM_LEFT: string; MIDDLE_LEFT: string; }>'.

Please move to this to discussion if this invalid issue.

aabounegm:
The `position` prop of the `<SnackbarContainer>` component was mistakenly typed as the object exported in `snackbar-positions.js`, instead of its values. Additionally, the type of that object was inferred to be
```ts
type SnackbarPositions = {
  TOP_LEFT: string;
  // ...
}
```
instead of
```ts
type SnackbarPositions = {
  TOP_LEFT: 'top left';
  // ...
}
```
so the `positions` fix alone is not enough. The function introduced to work around this issue may seem redundant, but IMO it is better than duplicating the entire object in an `@type` annotation (at least until the linked issue gets resolved).

Fixes #256


attractions/rollup.config.js
```diff
@@ -23,6 +24,20 @@ function prependTagOption(exceptions = []) {
   };
 }
 
+/**
+ * Recursively traverses all subdirectories of a given directory, listing all files it contains
+ * @param {string} dir Path to the directory whose files to list
+ * @returns {string[]} The list of files
+ */
+function getFiles(dir) {
```
illright, line 32: Maybe this is a better fit for a `utils` file?

aabounegm: I mean, it's only useful here for the rollup task. Functions in the `utils` are suitable for the (web) app itself, which doesn't have node. Who would need to list files in a directory as part of their build process if not for a use case as specific as this one?

illright (approved):

PR #276, opened 2021-07-18
Move to `pnpm`

illright:
[`pnpm`](pnpm.io) is the new (sort of) kid on the block, outperforming Yarn and npm 7 in most of the scenarios as well as being more efficient in terms of storage. It also offers strictness of `node_modules` usage, so that you only have access to your direct dependencies.

Its CLI is a kind of hybrid of npm and Yarn:
```
pnpm i  # supports shorthands of npm
```

```
pnpm add attractions --save-dev  # uses 'add' but also 'save-dev'
```

```
pnpx husky-init  # also offers a runner like npx
```

Yay or nay?


aabounegm (approved):
I already installed yarn just for this project, so I don't see why I can't switch again :)

illright:
With this latest commit I refactored the linting scripts for readability and replaced single quotes for double quotes, so if we merge this, we can close #278

aabounegm (commented):
Due to the inherit differences between CMD and PowerShell, it seems impossible to support both `script-shell` configs together. Therefore, I think it is best to just stick with the default one (CMD), mainly because many dependencies have pre/post-install scripts that are designed for Linux but also work on CMD (but not PowerShell). I believe supporting either of them does not break compatibility with most Linux shells, so the implications are the following:
- Use `&&` instead of `;`
- Use `"` instead of `'`
- No need to wrap `--` in quotes anymore (which you didn't, but just saying)

Also, I don't think this entirely replaces #278 as there are still outdated deps (reported by `pnpm outdated`), and the `package.json` in sub-projects are not updated in this PR

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented here.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
```
aabounegm, line 7: It should be `## [Unreleased]` to avoid conflicts with other branches

aabounegm (approved):

aabounegm (approved):

PR #277, opened 2021-08-06
Move the `maxReachedTooltip` variable to a prop.

### LINKED ISSUE #13: Allow customizing the max-reached label for checkbox groups
As of now, it is fixed to `Can only select {} values.`. This could be customizable with a prop to allow for i18n.

aabounegm:
This leads to a loss in reactivity for the default value, but there is no neat way (I know of) that allows customizing it while maintaing reactivity of the default value.

Closes #13


illright:
I'm thinking of maybe moving this to a named slot? That way we retain reactivity of the default value and consistency with the way this stuff is customized in `Autocomplete`.

aabounegm:
That's not possible because we use that value in the [HTML title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title). We could replace it with a `Popover` (acting like a tooltip), but I don't prefer that option (feels like an overkill).

illright:
We could do something like:
```js
export let maxReachedTooltipOverride = null;
$: maxReachedTooltip = maxReachedTooltipOverride ?? `...`;
```

aabounegm:
I thought about that option but didn't think it was neat (felt like a hack). But now that I see it written, it may not be that bad; seems decent enough.

attractions/checkbox/checkbox-group.svelte
```diff
@@ -51,9 +51,15 @@
    * @type {number}
    */
   export let max = Infinity;
+  /**
+   * The text displayed in the tooltip when hovering over the checkboxes
+   *   after the maximum allowed selection has been reached.
+   */
```
aabounegm, line 57:
```suggestion
   *   after the maximum allowed selection has been reached.
   * Leaving it as `null` defaults to `` `Can only select ${max} value${s(max)}.` ``
   * @type {string | null}
   */
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/) and this
 
 ### Added
 
+- `maxReachedTooltip` prop for the `CheckboxGroup` and `CheckboxChipGroup` component.
```
aabounegm, line 11:
```suggestion
- `maxReachedTooltip` prop for the `CheckboxGroup` and `CheckboxChipGroup` components.
```

aabounegm: And maybe a few words about what it does, like _"... to customize the hover text when `max` items have been selected"_

illright (approved):

PR #278, opened 2021-08-06
Update deps

aabounegm:
Update dependencies (using `yarn upgrade` - so minor, nonbreaking updates only).
Update all `package.json` scripts to use double quotes for better compatibility with other shells (mainly CMD on Windows, which is still the default `script-shell` for some reason)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented here.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


illright (approved):
What's this "Open in VS Code" feature? Is there a way to customize the "setup" like specifying a command to run after cloning or is it a mere cloning button?

aabounegm:
It is just a badge that helps new contributors (especially ones with small quick fixes) access the repo code faster using the relatively new [Remote Repositories](https://code.visualstudio.com/blogs/2021/06/10/remote-repositories) feature (introduces in the [May 2021 update](https://code.visualstudio.com/updates/v1_57#_remote-repositories)), or clones it in a dedicated Docker container. In both cases, the repo is opened directly in VS Code and contributors can get to the source code faster.
I believe it does not currently support any kind of customization.

aabounegm:
Already covered in #276

PR #280, opened 2021-08-06
calendar/disables-dates

### LINKED ISSUE #34: Implement disabling certain days for the calendar
So that one could specify an array of days or ranges (unbounded too)

aabounegm:
Closes #34

### DIFF
--- a/attractions/date-picker/calendar.scss
+++ b/attractions/date-picker/calendar.scss
@@ -44,9 +44,6 @@ $side-padding: 1rem;
       z-index: 2;
       width: $day-size;
       height: $day-size;
-      &:disabled {
-        color: vars.$disabled;
-      }
     }
 
     &.today > :global .btn {
@@ -64,6 +61,22 @@ $side-padding: 1rem;
 
     &.outside > :global .btn {
       color: vars.$disabled;
+    }
+
+    &.disabled > :global .btn {
+      color: vars.$danger;
+      position: relative;
+
+      &::before {
+        position: absolute;
+        content: '';
+        height: 1px;
+        width: 50%;
+        top: 50%;
+        left: 50%;
+        transform: translateX(-50%) rotate(-30deg);
+        background: vars.$danger;
+      }
     }
 
     &.in-range::before {
--- a/attractions/date-picker/calendar.scss
+++ b/attractions/date-picker/calendar.scss
@@ -44,6 +44,9 @@ $side-padding: 1rem;
       z-index: 2;
       width: $day-size;
       height: $day-size;
+      &:disabled {
+        color: vars.$disabled;
+      }
     }
 
     &.today > :global .btn {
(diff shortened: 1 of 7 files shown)


attractions/date-picker/calendar.scss
```diff
@@ -44,6 +44,9 @@ $side-padding: 1rem;
       z-index: 2;
       width: $day-size;
       height: $day-size;
+      &:disabled {
+        color: vars.$disabled;
```
illright, line 48: Maybe also `text-decoration: strikethrough`? How would that look like?

attractions/utils/datetime-utils.js
```diff
@@ -228,6 +238,15 @@ export function getCalendar(month, year, firstWeekday) {
       week.push({
         value: new Date(dayCursor.valueOf()),
         outside: dayCursor.getMonth() !== month,
+        // TODO: is this fine or should it be moved to a separate function?
```
illright, line 241: Yeah, I think it should be moved into a separate function

illright (changes requested):

illright (approved):

illright:
Apparently, I can't request your review on your own PR, but do review, please, before we get this merged

attractions/utils/datetime-utils.js
```diff
@@ -206,14 +215,35 @@ export function datesLessEqual(date1, date2) {
   );
 }
 
+/**
+ * Checks if the date is included in the given array of dates or ranges.
+ * @param {Date} date The date object, which inclusion to check
```
aabounegm, line 220:
```suggestion
 * @param {Date} date The date object whose inclusion is to be checked
```

attractions/utils/datetime-utils.js
```diff
@@ -206,14 +215,35 @@ export function datesLessEqual(date1, date2) {
...
+ * @param {Array<Date | { start?: Date; end?: Date; }>} dateRanges The set of dates
+ * @returns {boolean}
+ */
+export function dateIncluded(date, dateRanges) {
```
aabounegm, line 224: I was thinking since this got refactored to its own function, its call should be decoupled from `getCalendar`, and instead be composed in the Calendar component.
At least that was before you used `day.disabled` in two props instead of one, so I'm no longer sure.

aabounegm: Also, should its documentation be added to the utilities docs or do we only document ones directly exported from `utils/index`?

aabounegm (commented):
Also cannot approve my own PR, but consider it approved

attractions/utils/datetime-utils.js
illright: I'd prefer we only document the direct exports, and consider the rest to be internal utilities

illright (approved):

PR #287, opened 2021-08-19
Breadcrumbs

### LINKED ISSUE #157: Question about Breadcrumbs and Navigation component?
@illright I notice that you have the breadcrumbs component on the doc site but is not available from the library. Same to the navigation bar component; is there a particular reason that not supporting it?

Many thanks for your time.

aabounegm:
Implements the `Breadcrumbs` component, taken from the docs and generalized to allow any number of segments, as well as customizing the segments and separators.

Closes #157


illright:
The docs don't seem to work when you click "docs" from the homepage

aabounegm:
And that's why I hate `null`

docs/src/routes/docs/components/breadcrumbs.svx
```diff
@@ -0,0 +1,101 @@
...
+
+### `separator` slot
+
+If a string separator is not enough, use this slot to add a custom component separator (and ignore the prop).
+
+## SCSS Variables {#scss-variables}
+
+| Name | Description | Default |
+| ---- | ----------- | ------- |
+| **`$thin-font-weight`** | Description. | `300` |
```
illright, line 101:
```suggestion
| **`$thin-font-weight`** | The font size of the separator and the segments without `href`. | `300` |
```

attractions/breadcrumbs/breadcrumbs.scss
```diff
@@ -0,0 +1,13 @@
+@use '../_variables' as vars;
+
+nav {
+  display: flex;
+  align-items: center;
+  flex: 1;
+  margin: 0 2em;
+  font-weight: vars.$thin-font-weight;
+
+  :global(.node) {
```
illright, line 10: This might leak inside the custom segments
```suggestion
  & > :global .node {
```

illright (changes requested):

docs/src/routes/docs/components/breadcrumbs.svx
aabounegm: Whoops. That's what happens when you just copy the template :D

attractions/breadcrumbs/breadcrumbs.scss
aabounegm: That was kinda the point, to be able to give the `node` class to a slot

illright: Yeah, this still provides this option, but only for direct children, which is where you'd want it

illright (approved):

PR #289, opened 2021-08-21
Create a `Table` component

### LINKED ISSUE #255: Tables
First and foremost, thanks for sharing this library. great job.

I see that you have a component for pagination, how would you use it with a table?
  Implicitly, I am also asking for the table component which I couldn't find.

Thank you

aabounegm:
The component was directly taken from the docs, with a few added props and slots.

I am also thinking that:
1- slots should be inside the `th`/`tr`/`td`, not replace them (to preserve proper semantics and avoid mistakes by forgetting to add the wrapping element)
2- the `header` slot should be like the `item` one in that it replaces just one header (which it takes as a prop) and not the entire row. Or should we make both slots?

And I feel that the wording in the docs can be improved a bit.

Closes #255

### DIFF
--- a/attractions/table/table.scss
+++ b/attractions/table/table.scss
@@ -1,0 +1,72 @@
+@use 'sass:color';
+@use 'node_modules/attractions/_variables' as vars;
+
+table {
+  border-spacing: 0;
+  border-collapse: separate;
+  display: block;
+  overflow-x: auto;
+  margin-bottom: 1em;
+
+  th {
+    text-align: start;
+    padding: 0.5em 1em;
+    &.center {
+      text-align: center;
+    }
+    &.end {
+      text-align: end;
+    }
+  }
+
+  tr {
+    &.alternating {
+      &:nth-child(even) {
+        background: color.adjust(vars.$main, $alpha: -0.97);
+      }
+    }
+
+    &:first-child :global {
+      > td:first-child {
+        border-top-left-radius: vars.$table-radius;
+      }
+
+      > td:last-child {
+        border-top-right-radius: vars.$table-radius;
+      }
+    }
+
+    &:last-child :global {
+      > td {
+        border-bottom-width: 1px;
+      }
+
+      > td:first-child {
+        border-bottom-left-radius: vars.$table-radius;
+      }
+
+      > td:last-child {
+        border-bottom-right-radius: vars.$table-radius;
+      }
+    }
+  }
+
+  td {
+    padding: 1em;
+    border: 0 solid color.adjust(vars.$main, $alpha: -0.75);
+    border-width: 1px 0 0 1px;
+    line-height: 1.3em;
+
+    text-align: start;
+    &.center {
+      text-align: center;
+    }
+    &.end {
+      text-align: end;
+    }
+
+    &:last-child {
+      border-right-width: 1px;
+    }
+  }
+}
--- a/attractions/table/table.scss
+++ b/attractions/table/table.scss
@@ -11,9 +11,11 @@ table {
   th {
     text-align: start;
     padding: 0.5em 1em;
+
     &.center {
       text-align: center;
     }
+
     &.end {
       text-align: end;
     }
@@ -58,9 +60,11 @@ table {
     line-height: 1.3em;
 
     text-align: start;
+
     &.center {
       text-align: center;
     }
+
     &.end {
       text-align: end;
     }
--- a/attractions/table/table.scss
+++ b/attractions/table/table.scss
@@ -28,7 +28,7 @@ table {
       }
     }
 
-    &:first-child :global {
+    &:first-child {
       > td:first-child {
         border-top-left-radius: vars.$table-radius;
       }
@@ -38,7 +38,7 @@ table {
       }
     }
 
-    &:last-child :global {
+    &:last-child {
       > td {
         border-bottom-width: 1px;
       }
--- a/attractions/table/table.svelte
+++ b/attractions/table/table.svelte
@@ -38,42 +38,36 @@
 </script>
 
 <table {...$$restProps}>
-  <slot>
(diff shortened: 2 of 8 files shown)


attractions/table/table.svelte
```diff
@@ -0,0 +1,75 @@
...
+  <slot>
+    <slot name="header">
+      <thead>
+        <tr>
+          {#each headers as header (header.value)}
+            <th
+              class:center={header.align === 'center'}
+              class:end={header.align === 'end'}
+            >
+              <Label>{header.text}</Label>
```
illright, line 45: Yeah, maybe we should put the slot here and send through the slot prop `headerItem`?
```suggestion
            <slot name="th" headerItem={header}> 
              <Label>{header.text}</Label>
            </slot>
```

attractions/table/table.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  td {
+    padding: 1em;
+    border: 0 solid color.adjust(vars.$main, $alpha: -0.75);
+    border-width: 1px 0 0 1px;
+    line-height: 1.3em;
+
+    text-align: start;
+    &.center {
```
illright, line 61:
```suggestion

    &.center {
```
And the same for the other rules

attractions/table/table.svelte
```diff
@@ -0,0 +1,75 @@
...
+  export let headers = [];
+  /**
+   * The actual data displayed in the table rows.
+   * The keys of the objects have to belong to the `value`s of the headers.
+   * @type {Array<Item>}
+   */
+  export let items = [];
+
+  /**
+   * Rows should have alternating background colors.
```
illright, line 29:
```suggestion
   * Whether the table rows should have alternating background colors.
```

attractions/table/table.svelte
```diff
@@ -0,0 +1,75 @@
...
+   *  align?: "start" | "center" | "end";
+   * }} Header
+   * @typedef {{[value: string]: string | number}} Item
+   *
+   * @slot {{ header: Header; item: Item }} item
+   * @slot {{ item: Item }} row
+   */
+
+  /**
+   * The header row of the table. `text` is the text to display, and `value` is used to link the row data.
```
illright, line 17: I don't remember if our docs support such elaborate markup, but I hope so
```suggestion
   * The labels for the column headers.
   *
   * Each column header is an object with the following fields:
   * - `text` is the text to display;
   * - `value` is the name of the field to take from each object in `items` to put in this column;
   * - `align` is the alignment of the text in the column (`"start" | "center" | "end"`).
```

illright (changes requested):
As for the slots, yeah, let's move them inside `th` and `td`

attractions/table/table.svelte
aabounegm: > our docs

This is part of the component's JSDoc, not our `svx` documentation. Did you mean to suggest this there?

attractions/table/table.svelte
aabounegm: Do you mean instead of the `header` slot or in addition to it?

illright: Instead of it

attractions/table/table.svelte
illright: Oh, right, I forgot that we don't use the JSDocs in the actual documentation. Well, I think it works in both

attractions/table/table.scss
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  tr {
+    &.alternating {
+      &:nth-child(even) {
+        background: color.adjust(vars.$main, $alpha: -0.97);
+      }
+    }
+
+    &:first-child :global {
```
illright, line 31: Why are these styles global?

attractions/table/table.scss
```diff
@@ -0,0 +1,76 @@
...
+      > td:first-child {
+        border-top-left-radius: vars.$table-radius;
+      }
+
+      > td:last-child {
+        border-top-right-radius: vars.$table-radius;
+      }
+    }
+
+    &:last-child :global {
```
illright, line 41: And here as well

docs/src/routes/docs/components/table.svx
```diff
@@ -0,0 +1,104 @@
...
+
+<Showcase>
+  <div slot="showcase" class="padded">
+    <Table {headers} {items} />
+  </div>
+  <div slot="source">
+
+```svelte
+<script>
+const headers = [
```
illright, line 43: Let's indent this 2 spaces

attractions/table/table.svelte
```diff
@@ -0,0 +1,74 @@
+<script>
+  import Label from '../typography/label.svelte';
+
+  /**
+   * @typedef {{
+   *  text: string;
+   *  value: string;
+   *  align?: "start" | "center" | "end";
```
illright, line 8: I just realized, I don't really like `start` and `end`:
* `left` and `right` would fit better with `center`
* what's a start for me is the end for you, Mr. RTL

illright (changes requested):

attractions/table/table.scss
aabounegm: idk. I copied them like that from the docs

attractions/table/table.svelte
aabounegm: > what's a start for me is the end for you, Mr. RTL

that is precisely the reason they are named as such. `start` and `end` were added to [`text-align`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align) in CSS3 to adapt to the language direction. `right` and `left` are rigid, while `start` and `end` respect the [`direction`](https://developer.mozilla.org/en-US/docs/Web/CSS/direction) setting

illright (approved):

PR #302, opened 2021-09-12
Convert the root package configuration to YAML

illright:
I'm not implying anything but this... looks kinda neat


illright:
Ah crap, Prettier rewrote it slightly

illright:
Alright, now it's not too neat. Just take a look at what's possible and feel free to close this

aabounegm (commented):
I'm sorry, WHAT??
Do package managers even support this? It seems like only `pnpm` does.
Like, why would you even wanna? And doesn't this break the functionality where we import the package json to extract the version number or other information? It definitely breaks compatibility with every package manager as they all read the `package.json` of all deps to run any `postinstall` scripts and so on. And why did you even do this change just for the root package and not the library or docs?

Just, no)
speechless

PR #309, opened 2021-10-09
Slider-changelog

aabounegm:


RikuVan:
Ah didn't see all the github notifications until today, so i missed the chance to comment. But maybe that is better as already the PR hung around so long, I had kind of forgotten about it :). I am glad something came of it eventually.

Issue #293, opened 2021-08-22
Change the `Dropdown` chevron animation


illright:
Change the `Dropdown` chevron animation to a `scaleY(-1)` for `150ms`. Looks nicer

Th1nkK1D:
Hi, I created a PR for this issue: #310

PR #310, opened 2021-10-10
Update dropdown chevron transition

Th1nkK1D:
As discussed in issue #293 I updated the dropdown chevron transition
- Use transform scaleY
- Change the duration to 150ms

Not sure if I should update the CHANGELOG.md as well. Please let me know if it's needed.


illright:
Hi! Thank you for your contribution. Yes, this change should be reflected in the changelog

Th1nkK1D:
The changelog is updated! 😄

PR #313, opened 2021-10-12
`closeOnClick` prop for the `DatePicker`

### LINKED ISSUE #307: Add closeOnClick Prop to DatePicker
Is it possible to add a prop which closes the dropdown calendar of the DatePicker automatically when a date is selected?

aabounegm:
I implemented it like in the referenced issue, but I feel like `closeOnSelected` would be a more appropriate name (especially to account for the `range` date picker). What do you think?

Resolves #307


illright:
I'd rather we call it `closeOnSelection`. But now that I think about it, I kind of wouldn't want to see a prop like that in Svelte code. It's kind of this huge monstrosity that doesn't look good without a `={true}` after the name. From this perspective, I'd rather call the prop `autoclose`. Thoughts?

aabounegm:
@illright For a boolean prop with a default value of false, I don't think it matters if it looks good with an = after the name, since you almost never it. You usually either provide the prop name only or you don't provide anything.
autoclose feels a bit ambiguous because it is not clear what this "auto" is based on (the selection, as made clear by the other name).
I'd prefer closeOnSelection

PR #317, opened 2021-10-14
Fix the handing of SCSS by ESLint in Svelte files

illright:


aabounegm:
I meant setting it to `true` unconditionally:

`'svelte3/ignore-styles': true`

PR #318, opened 2021-10-31
`Dropdown` automatic positioning

### LINKED ISSUE #294: Implement automatic positioning of the `Dropdown`
Implement automatic positioning for the `Dropdown` (and hence the `DatePicker` and `TimePicker`).

Th1nkK1D:
As discussed in #294, here is the implementation detail

## Changes
- Add `horizontalAlignment` prop with the following possible values: auto-start (default), auto-end, start, end
- Add `verticalAlignment` prop with the following possible values: auto-top, auto-end, bottom, top
- The auto- option is the automatic positioning with the preferred alignment when either is available.
- Define and export two enums (frozen objects, since we don't use TS) from the Dropdown component that define these values
- Mark `top` and `right` props as deprecated, and will be removed in the future.
- Update changes to doc

---
Maintainer's edit: closes #294


docs/src/routes/docs/components/dropdown.svx
```diff
@@ -36,11 +36,21 @@ Toggleable, contextual overlays for displaying additional details or actions. Mu
     </DropdownShell>
     <DropdownShell let:toggle>
       <Button on:click={toggle}>
-        look up there
+        Look up there
```
illright, line 39: That was intentional, currently the styleguide of Attractions is to write buttons in lowercase. That is strange, however, and will change after the GrEaT reDeSIgN.

Th1nkK1D: HAHAHA OK, I didn't notice that. I Will change it back

illright (commented):
I'm not too keen on exposing the DOM element of the dropdown shell, as well as making it a store. Maybe we could instead expose a callback that would query the bounding rect of the shell element? From the looks of it, that's all we need the DOM element for.

It was the first thing that caught my eye, I'll comment on the rest after we settle with this

illright:
Oh and did I mention what a great job you've done? Thanks a lot!

Th1nkK1D:
> I'm not too keen on exposing the DOM element of the dropdown shell, as well as making it a store. Maybe we could instead expose a callback that would query the bounding rect of the shell element? From the looks of it, that's all we need the DOM element for.
> 
> It was the first thing that caught my eye, I'll comment on the rest after we settle with this

Great point. Fixed!



> Oh and did I mention what a great job you've done? Thanks a lot!

Thanks! It was more complicated than I thought but it feel great to contribute to the community!

attractions/dropdown/dropdown-shell.svelte
```diff
@@ -49,6 +55,15 @@
       : document.removeEventListener('keydown', handleKeyPress));
 
   const dispatch = createEventDispatcher();
+
+  const isDropdownOpen = writable(open);
+  const getDropdownShellBoundary = writable(() => {});
```
illright, line 60: Perhaps this would suffice?
```suggestion
  function getDropdownShellBoundary() {
    if (self == null) {
      return null;
    }
    
    return self.getBoundingClientRect();
  }
```

Th1nkK1D: Yup I'm overthinking it LOL. I make your suggestion shorter:

```typescript
const getDropdownShellBoundary = () => self?.getBoundingClientRect();
```

attractions/dropdown/dropdown.svelte
```diff
@@ -13,9 +37,78 @@
...
+    switch (horizontalAlignment) {
+      case HorizontalAlignment.end:
+        return true;
+      case HorizontalAlignment.start:
+        return false;
+      case HorizontalAlignment.autoEnd:
+        return width <= left;
+      default:
+        // auto-start by default
+        return width > window.innerWidth - right;
```
illright, line 88: I think this should be counted from the left, because when the dropdown is start-aligned, it spans the entire width of the dropdown shell and the rest of the page on the right.
```suggestion
        return width > window.innerWidth - left;
```

illright (changes requested):
In the docs, the showcase, the topmost dropdown opens outside of the showcase and gets cut out. Could you please alter the styles of this docs page to prevent that from happening?

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -81,8 +101,28 @@ Toggleable, contextual overlays for displaying additional details or actions. Mu
 
 | Name | Default | Type | Description |
 | ---- | ------- | ---- | ----------- |
-| **`right`** | `false` | `boolean` | Aligns the dropdown content with the right edge of the shell. |
-| **`top`** | `false` | `boolean` | Aligns the dropdown content such that it opens to the top of the shell, extending up. |
+| **`horizontalAlignment`** | `auto-start` | `auto-start`, `auto-end`, `start`, `end` | Dropdown content alignment on horizontal axis. |
```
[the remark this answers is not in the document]
Th1nkK1D:
```md
`auto-bottom` \| `auto-top` \| `bottom` \| `top`
```
did the trick

attractions/dropdown/dropdown-shell.svelte
```diff
@@ -49,6 +55,14 @@
       : document.removeEventListener('keydown', handleKeyPress));
 
   const dispatch = createEventDispatcher();
+
+  const isDropdownOpen = writable(open);
+  const getDropdownShellBoundary = () => self?.getBoundingClientRect();
```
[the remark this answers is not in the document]
Th1nkK1D: I changed it to

```ts
  const getDropdownShellBoundary = () => self && self.getBoundingClientRect();
```
should be OK right?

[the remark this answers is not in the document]
illright: Yeah, I think so

Th1nkK1D:
> In the docs, the showcase, the topmost dropdown opens outside of the showcase and gets cut out. Could you please alter the styles of this docs page to prevent that from happening?

All fixed!

attractions/dropdown/dropdown-context-key.js
```diff
@@ -0,0 +1,2 @@
+export const CONTEXT_GET_DROPDOWN_SHELL_BOUNDARY = 'getDropdownShellBoundary';
```
illright, line 1: I don't think we need the `CONTEXT_` prefix. Also, it's preferable to use `{}` as a value of a context key because no two objects are referentially equal to each other, so we don't risk collisions with other Svelte code. Also I think context keys look better in simple camelCase.
```suggestion
export const getDropdownShellBoundaryKey = {};
```

attractions/dropdown/dropdown.svelte
```diff
@@ -13,9 +22,78 @@
...
+    const { width } = dropdownBound;
+    const { left } = dropdownShellBound;
+
+    switch (horizontalAlignment) {
+      case HorizontalAlignment.END:
+        return true;
+      case HorizontalAlignment.START:
+        return false;
+      case HorizontalAlignment.AUTO_END:
+        return width <= left;
```
illright, line 70: Then this should be `right`
```suggestion
        return width <= right;
```

attractions/dropdown/dropdown.svelte
```diff
@@ -13,9 +22,78 @@
...
+      default:
+        // auto-start by default
+        return width > window.innerWidth - left;
```
illright, line 73: Now I'm also contemplating if we should allow specifying which element to use as the boundary. For instance, if we have an element with `overflow: hidden`, we'd want to position the dropdown within the bounds of that element. Can we do that by just passing a DOM node of the respective element, defaulting to `window`?

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -101,6 +123,30 @@ Falsy values passed to classes will be disregarded.
...
+
+The following objects are available as named exports in 'attractions/dropdown'.
+
+
+#### HorizontalAlignment
+Dropdown content alignment on horizontal axis.
+
+| Field | Description |
+| ---- | ------- |
+| **`AUTO_START`** | Align content based on available space, preferable start. |
```
illright, line 136: I think this explains the `start`/`end` a bit more clearly:
```suggestion
| **`AUTO_START`** | Align the dropdown based on available space around, preferably, at text start (left edge for LTR languages). |
```

docs/src/routes/docs/components/dropdown.svx
```diff
@@ -101,6 +123,30 @@ Falsy values passed to classes will be disregarded.
 
 The content of the dropdown.
 
+### Related Objects
+
+The following objects are available as named exports in 'attractions/dropdown'.
+
+
+#### HorizontalAlignment
+Dropdown content alignment on horizontal axis.
```
illright, line 132: I suppose it's not really the content that we're aligning, but the dropdown itself: 
```suggestion
Dropdown alignment along the horizontal axis.
```

attractions/dropdown/dropdown-alignment.js
```diff
@@ -0,0 +1,36 @@
...
+  AUTO_END: 'auto-end',
+  START: 'start',
+  END: 'end',
+};
+
+/**
+ * This is needed to improve type inference in JSDoc, as it doesn't yet support `as const` assertions.
+ * See [microsoft/TypeScript#30445](https://github.com/microsoft/TypeScript/issues/30445) for more details.
+ * @type {{
+ * AUTO_BOTTOM: 'auto-bottom',
```
illright, line 22: Could you please indent these lines? And the ones above, with 2 spaces from the text, currently it's just one.
```suggestion
 *   AUTO_BOTTOM: 'auto-bottom',
```

illright (changes requested):

attractions/dropdown/dropdown.svelte
Th1nkK1D: Sorry, I was very busy in the previous months.

For this feature, It is quite an edge case and I'm not sure if it will be worth introducing the complexity to the code base. **window** use `innerWidth` and `innerHeight` while **HTMLElement** use `clientHeight` and `clientWidth`. And when we use HTMLElement, which is mostly smaller than the window, then its position must be included in the calculation as well.

aabounegm:
Hmm. I'm not sure what I messed up by merging from `main`, but it shows the same commits that already exist in `main` (and their changes in the Files tab) for some reason.

aabounegm:
Ok, hopefully that fixes it. Sorry for the mess up with the merge yesterday 😁 .

aabounegm (approved):

illright (approved):

illright:
Phew, finally. Thanks a lot, @Th1nkK1D!

attractions/dropdown/dropdown.svelte
illright: Alright, we might add it in the future if there will be any requests for it.

PR #319, opened 2021-11-03
Change from absolute color to scss varaible

Chaostheorie:
Minor fix to adapt `card`'s background to use the theme background variable instead of using an absolute value.


illright:
Sounds good!

illright (approved):

Chaostheorie:
Thank you for the quick merge.

PR #334, opened 2022-01-03
`class` prop type

### LINKED ISSUE #333: Can't pass class props to some components
Hi, great work for that library!
I saw that I can't pass class prop to Chip component also if it's defined but I don't found in the soruces the cause 🤦🏽‍♂️
image

aabounegm:
After the last update in Sveld, it became possible to give `@type`s to props defined via a rename export, enabling us to give a type to the `class` prop.

Caveats:
- In components where `{...$$restProps}` is used, an error will occur in the declaration file similar to the following:
```
Interface 'BadgeProps' incorrectly extends interface 'HTMLAttributes<HTMLDivElement>'.
  Types of property 'class' are incompatible.
    Type 'string | false | null | undefined' is not assignable to type 'string | null | undefined'.
      Type 'false' is not assignable to type 'string | null | undefined'
```
However, this should ultimately not affect the end user. On the contrary, allowing `false` values is a utility for users.

Additionally, this fixes many more type errors, which brings the actionable errors of `svelte-check` to a minimum (there are still a couple errors that we can fix, while the rest is out of our hand).
Closes #333


attractions/time-picker/time-picker.svelte
```diff
@@ -220,7 +226,10 @@
             on:click={() =>
               setHours(
                 hour +
-                  12 * (f12hours && (currentAmPm === 'PM') ^ (value === 12))
+                  12 *
+                    Number(
+                      (f12hours && currentAmPm === 'PM') !== (value === 12)
```
aabounegm, line 231: In particular, this is one the errors reported by `svelte-check` that makes sense. `value` is of type `Date | null`, which means that the condition `value === 12` will always be `false`

attractions/checkbox/checkbox.svelte
```diff
@@ -70,8 +71,8 @@
     {disabled}
     on:change={e =>
       dispatch('change', {
-        value: e.target.value,
-        checked: e.target.checked,
+        value: e.currentTarget.value,
```
illright, line 74: Why this? As far as I know, `currentTarget` is a more problematic field than `target` in regards to browser compatibility

attractions/file-input/file-input.svelte
```diff
@@ -27,7 +28,7 @@
   export let vertical = false;
   /**
    * The user's selection. If `multiple` is `false`, the value is an actual `File` object, not a one-element `FileList`, as opposed to the native `<input type="file">`.
-   * @type {File | FileList | null}
+   * @type {File | File[] | null}
```
illright, line 31: Again, are you sure about this? I see from the initial value that this is correct, but in the usage later on?

attractions/time-picker/time-picker.svelte
illright: Yeah, that's probably a bug)

attractions/utils/datetime-utils.js
```diff
@@ -2,11 +2,17 @@ const numberRegex = /\d+/g;
 const formatSpecifierRegex = /%[HMSpPdmyY%]/g;
 const daysInWeek = 7;
 
+/**
+ * @template T1
+ * @template T2
+ * @typedef {T1 extends null ? T1 : T2 extends null ? T1 : T2} NullFirst
```
illright, line 8: Not sure I can work out from this name what `NullFirst` actually is

illright (commented):
A few questions

attractions/checkbox/checkbox.svelte
aabounegm: Actually, according to caniuse.com, [`currentTarget`](https://caniuse.com/mdn-api_event_currenttarget) is supported even more than [`target`](https://caniuse.com/mdn-api_event_target), with the only difference being in IE 6-8 (so who gives a damn?).
As for why, there is a minor difference between both. `target` is the element where the event happened, while `currentTarget` is the element to which the listener is attached. They are the same in our case, but Svelte seems to augment the type for `currentTarget` so that it is not nullable and has the correct element type.

attractions/file-input/file-input.svelte
aabounegm: No, I'm not so sure about this one, it just seemed to silence the errors in our code. As I said, not everything is fixed, but this here has a bigger problem than types, imo. It is sometimes a `File[]` (whenever we set it manually), but other times a `FileList` (when the `<input>` is updated), and neither of them is assignable to the other.

attractions/time-picker/time-picker.svelte
aabounegm: So should I just remove the `!== (value === 12)` part? Also, is there a specific reason it was `^` instead of `!=` previously?

attractions/utils/datetime-utils.js
aabounegm: It's to properly type functions with the following guard:
```ts
if (source == null || destination == null) {
  return source;
}
```
that do not return `null` otherwise. The functions are sometimes called with non-nullable arguments, but the return type was still `Date | null`, which caused errors when using the returned value as non-nullable. This type better represents the implementation.

You can read like so: if the first parameter is `null`, then the function returns `null`. Otherwise, if the second parameter is `null`, then we return the first parameter (as non-nullable). Otherwise, we return the \[type of the\] second parameter (also non-nullable).

I agree the name is not very descriptive, but suggestions are welcome.

attractions/time-picker/time-picker.svelte
illright: You can remove it, but I suggest you don't, since it's kind of out of scope for this PR. Instead we should follow up with a PR that fixes other issues in our code, apart from classes.

Answering the question about `^` – no, I'm pretty sure there was no reason

attractions/checkbox/checkbox.svelte
illright: Huh. Well, alright, as long as nothing breaks (though I'm kind of scared that it will, maybe not for this component but for another)

attractions/file-input/file-input.svelte
illright: I suggest we revert this change for now, lest anything gets out of hand, and come back to it in a separate PR

attractions/utils/datetime-utils.js
illright: How about `NullLikeArguments` or `NullFromArguments`? This util is making my head spin, I too cannot express it in words

attractions/time-picker/time-picker.svelte
aabounegm: I mean, everything that doesn't involve adding `@type` before `export { _class as class }` is kinda out of scope already)
If you prefer, I can drop the relevant commit from this branch and move it along with the additional fixes in another PR, but they just felt like they belong since it's all about fixing type errors. Also, just removing `!== (value === 12)` doesn't feel like it deserves its own PR)

attractions/utils/datetime-utils.js
aabounegm: How about `PossiblyFilterNull`? This utility type is not really specific to **arguments** only

attractions/time-picker/time-picker.svelte
illright: > doesn't feel like it deserves its own PR

I feel like it does. There's a reason why this condition was there and this error might indicate that a bug had crept in our code. We need to consider it separately.

As for moving this commit to another branch – no, I think it's fine here. It's out of scope but it's not that out of scope :)

attractions/utils/datetime-utils.js
illright: No, I don't really like any of those names, now that  I think about it. And to be honest, even this utility is a very complicated little snippet if you're not in context of what's happening (i.e. remembering everything we've talked about here). So I think that the healthiest solution is just to make the return type `Date | null` and add appropriate guards. Could this mean we'll need a lot of guards? Yeah, but then maybe we should start considering refactoring this code

aabounegm: I don't think it's really the healthiest solution. Types are generally supposed to dictate what implementation is correct, or in this case (since types aren't a part of JavaScript) to document the implementation as much as possible. Writing `Date | null` would be an inaccurate representation of what the function actually does. Perhaps the implementation itself is a little twisted (partially due to the lack of null safety in the whole project) and can be improved, but it is what it is at the moment

illright: Alright, let's just go with this for now and then in the future we might refactor the code to remove the need for such complicated types altogether

illright (approved):

PR #347, opened 2022-06-22
Update typings for required props

aabounegm:
This is achieved simply through updating sveld to `v0.17`. The rest is very minor changes


illright (commented):
Unsure about that change from `target` to `currenrTarget`, why did you do it?

aabounegm:
Check the type annotation. This is the one Svelte actually assigns to the event type automatically. You can see it by explicitly writing a function literal in the handler (`e => handleInput(e)`) and hovering over `e`. I believe this is because `currentTarget` is guaranteed to not be `null` in this case. Also, to quote a [Stack Overflow answer](https://stackoverflow.com/a/10086501):
- `target` is the element that triggered the event (e.g., the user clicked on)
- `currentTarget` is the element that the event listener is attached to.

illright (approved):

PR #348, opened 2022-06-29
Support custom body in tables

mijorus:
Hi,
I have been using attractions this week but I quickly found that tables are pretty limited because by default you can't fully customize rows etc.

A quick fix might be this one, please note this is just POC code that I will improve further I you like the idea.

Basically tables would have a `<slot />` named body where we can inject a `<tbody />`  element formatted as we like.
Let me know what you think

### DIFF
--- a/attractions/table/table.scss
+++ b/attractions/table/table.scss
@@ -1,7 +1,8 @@
 @use 'sass:color';
 @use 'node_modules/attractions/_variables' as vars;
 
-table {
+table,
+.attractions-table {
   border-collapse: separate;
   border-spacing: 0;
   display: block;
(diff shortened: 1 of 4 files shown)


aabounegm:
The idea itself is valid, but the implementation can be improved. To begin with, I don't like the fact that a prop controls the usage of a slot and would prefer to have the `tbody` content be the fallback slot content. In other words, instead of
```svelte
{#if customBody}
  <slot name="body" />
{:else}
  <tbody>...</tbody>
{/if}
```
it would be better if it looked like this:
```svelte
<slot name="body">
  <tbody>...</tbody>
</slot>
```
without using a prop at all.

That also makes me think that it might be a good idea to add more slots for `head` and `row` for more granular control.

mijorus:
Do you think the slot should be wrapped inside a tbody already or that it would be unnatural for a user to generate a table without wrapping it ?

mijorus:
Do you know how I can test this package locally?
Every time I try to `npm install git://github.com/mijorus/attractions.git#custom-table-rows --save-dev` it errors out:

```
`npm ERR! Cannot read properties of undefined (reading 'spec')`
```

aabounegm:
> Do you think the slot should be wrapped inside a tbody already or that it would be unnatural for a user to generate a table without wrapping it?

I think it would make more sense to have it outside it so that usage would like as such:
```svelte
<Table>
  <tbody slot="body">...</tbody>
</Table>
```

> Do you know how I can test this package locally?

You can clone the repo locally (or open in GitHub Codespaces) and run `pnpm docs:dev`.
You couldn't install it from git with npm because the `package.json` in the repo root is not the same as the one the package uses, we are using a monorepo style. You will need to use [pnpm](https://pnpm.io/)

mijorus:
@illright I am sorry for the ugly solution, but it's the only one I have found

mijorus:
I feel like the latest commit is a much better implementation both for the user and it's easier to maintain

aabounegm:
I don't understand why you added `<style lang="scss" src="../../../../node_modules/attractions/table/table.scss"></style>` to the docs. The docs have their own styles and they should not affect the selectors of the original component. It is intentional that slots are to be styled by the containing component that fills them in, not the one in which they get inserted.
Also, can you please make sure the CI passes and maybe add some documentation to the slots section of the page, not just the example?

mijorus:
I want my table body to look just like a table body that you would create with the `item` prop, but the content of each box can be freely personalized (say I want a button as the last element of one of my rows)

In order to do this, I can define  my custom `tbody` element, which however will now look completely different from the rest of the UI; to avoid this issue, I can import the attractions' stylesheet into my component, so that the `table` can look right

aabounegm:
I understand, but that is not the common use-case for slots; they are typically used with custom styles. You may import the scss file into your component, but we wouldn't want to modify the component's original styles, or the docs for that matter.
For the example you described above, you might be better off using the `item` slot instead, and use the `header` and `item` slot props to identify which cell it is.
I'm still not opposed to adding more slots, I just want them to be generic and not tailored for one specific usage scenario

PR #312, opened 2021-10-12
Migrate the docs to SvelteKit

illright:
Closes #311. Work in progress.

Issue #181, opened 2020-12-11
Improvement: use css custom properties instead of scss variables for themeing


RikuVan:
Presently Attactions requires a dependency on Sass for overriding variables. Personally I have found this troublesome since I am not really interested in using SASS otherwise. I recently tried to change the build and this was immediately a blocking issue. If Attractions used custom css properties (variables) for overrides (which is compatible with SASS), this would remove some friction for users. More info here: https://sass-lang.com/documentation/breaking-changes/css-vars.

illright:
There are several reasons why Attractions is currently using SCSS:

1. I personally had doubts about bringing in CSS variables because this would cut the browser compatibility. Admittedly, this is not a large cut at all, moreover, I suspect that in some cases there might be other things that render the library incompatible with older browsers such as new JS syntax that might be untranspiled or something like that.

2. Attractions also makes use of some very handy Sass features such as style nesting and color computation. Switching to pure CSS would mean messier stylesheets and just plainly lack of some features like proper hover colors and so on.

The first point is not a deal-breaker, I suppose we could switch the library to use CSS variables and still be highly supported by browsers. The second point, however, is a present matter of discussion in the Svelte community, especially with desirable things like TypeScript. Currently there is no sensible way of shipping preprocessed Svelte components, however, there are [claims](https://github.com/sveltejs/svelte/issues/604#issuecomment-727818761) that it should be taken care of in the upcoming [SvelteKit](https://www.youtube.com/watch?v=qSfdtmcZ4d0&feature=youtu.be). They say that the SvelteKit isn't too far out, so perhaps our best bet would be to wait for its release and then take the steps to ship preprocessed Svelte source.

Until then, however, I suppose we have to force users to use SCSS. Can you elaborate more on why you think it's troublesome? I understand it's an extra step in installation and a worthless devDependency, but other than that, is there anything we could do in the library that would make installing SCSS less painful? Open to any suggestions

aabounegm:
About the very last sentence, I suppose we could make it a direct dependency instead of a dev one, but that only saves users the trouble of installing it manually, not of configuring it (which is more inconvenient, imo). Perhaps we can also depend on `svelte-preprocess` and expose an object with the preprocessing configuration already written for those who don't want to customize the variables?

RikuVan:
[...]

PR #349, opened 2022-07-05
SvelteKit migration

### LINKED ISSUE #311: Migrate to SvelteKit
Sapper is no longer actively developed and SvelteKit is a much better experience

Migration guide: https://kit.svelte.dev/migrating

aabounegm:
A previous PR (#312) was made with the target of rewriting the docs in SvelteKit. It quickly became outdated and was only targeting the docs, so this PR supersedes it since it started from a fresh SvelteKit template and targets the entire app (not just docs).

# Important tasks

The migration is large and some of the tasks may be split into other smaller PRs. The general outline of the tasks is as follows:

- [x] Move all files to the correct locations in SvelteKit convention, fixing imports and layout props accordingly
- [x] Remove Custom Elements support (SvelteKit is just not ready for them yet, and they weren't even working properly)
- [x] Replace any Sapper reference with SvelteKit
- [x] Fix GitHub Actions and pre-commit hooks
- [x] Investigate why some `:global` selectors are not working as they should
- [ ] Figure out what to do with theming and SCSS variables (related to #181 - should probably convert/open a discussion)

# Additional tasks

These are the nice-to-haves that theoretically shouldn't block the migration to SvelteKit, but also make use of some features that were not possible/easy before.

- [x] Migrate everything to TypeScript
- [x] Turn off router and/or hydration for pages that don't need them
      Hydration is required for all pages (at the very least, for the "Components" dropdown to work), and in that case there is no compelling reason to disable the router.
- [x] Remove deprecated interfaces (only `right` and `top` props in `Dropdown`)
- [ ] Update documentation
  - [ ] Theming
  - [ ] Migration
  - [ ] Installation
  - [ ] Changelog

# Blocking issues

- ~~The PR is currently blocked by sveltejs/kit#4528 and the `{base}/...` workaround is not practical. Some links cannot be transformed that way (ex: ones inside CHANGELOG.md). Since `<base href="/" />` is allowed (and actually fixes the problem), I'm wondering why `base: '/'` is not allowed.~~
  **Update**: This turned out to be feasible. I was initially under the impression that `kit.paths.base` would add the `base` tag as Sapper did, but that turned out to not be the case. It was definitely tiresome but possible to update all URLs everywhere.

# Other issues

- Forwarded slots (`<slot name="loading-options" slot="loading-options">`) are detected as taking a slot prop called `slot` in the generated types.
- Svelte Language Server (VS Code extension) does not understand import assertions (`import pkg from './package.json' assert { type: 'json' }`) even with latest Node.js installed
  **Update**: Doesn't matter any more since I moved the package import to `vite.config.js`

Closes #311


aabounegm:
Hello @metaory, sorry for the delays from my side, but I just finished my final exams (at uni) today. I plan to come back to this issue soon, but I was also hoping to get more feedback on #360 first

benmccann:
Is there anything you're blocked on in terms of getting this in?

I feel like this task has maybe become too large. I think migrating only the docs first as https://github.com/illright/attractions/pull/312 was doing might have been more approachable vs trying to boil the ocean. But given how much progress has been made with this approach, maybe it could just be merged with a few things broken and some cleanup can happen later?

aabounegm:
Well, the main issue is that both maintainers (me and @illright) got busy with other things in their lives, be it study or work or both. Still, I wanted to continue with the migration because it didn't seem so difficult. The main stopping point was #360, in which we decided to move away from Sass to CSS with variables (custom properties), but SCSS/CSS are not really my strong point. I did some work in the [sass-migration](https://github.com/illright/attractions/tree/sass-migration) branch but it's far from complete.

**TL;DR**: I need to muster up some energy to continue the CSS migration and the rest should be straightforward

PR #369, opened 2023-06-21
Prevent svelte compiler from throwing accessibility warnings where we only care about the click event

noosxe:
Svelte compiler complains about non-interactive elements with on:click handlers not having a keyboard alternative  for a11y purposes.

This change adds missing event handlers.

### DIFF
--- a/attractions/autocomplete/autocomplete-option.svelte
+++ b/attractions/autocomplete/autocomplete-option.svelte
@@ -33,6 +33,7 @@
   const dispatch = createEventDispatcher();
 </script>
 
+<!-- svelte-ignore a11y-click-events-have-key-events -->
 <li on:click|stopPropagation={e => dispatch('click', { nativeEvent: e })}>
   <button type="button">
     {#each markMatch(option.name) as part}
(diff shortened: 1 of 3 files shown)


aabounegm:
Thank you for this contribution, but are you sure that treating all keypresses equally is the right thing to do here? I think people generally expect mouse clicks to be equivalent to pressing only <kbd>Enter</kbd> and/or <kbd>Space</kbd>

noosxe:
@aabounegm thanks for your comment.

I dug a little bit deeper and decided to just prevent the compiler from throwing the errors for these cases.

There is no point in adding keyboard event handlers at all because the elements can't be navigated to using a keyboard anyway.
The "real" accessibility is handled by elements that are wrapped by the ones that were throwing warnings in case of the autocomplete-option and file-dropzone. In case of modal clicking on the backdrop to hide the modal is useful but doing the same with the keyboard is not because keyboard users can instantly jump to the close button itself.

aabounegm (approved):
That makes sense. Thank you!

